PR #16, opened 2015-05-19
Design prototype for section marking and removal

kbairak:
@lolletsoc take a look when you have some time and let me know what you think.

It will need some documentation before merging.


lolletsoc:
Looks good!

PR #19, opened 2015-05-26
Fixed coverage

lolletsoc:
@glezos 

Added scripts to deal with the `python setup.py test` command

Should fix coverage.


kbairak:
@lolletsoc we won't know for sure if we don't merge, right?

PR #20, opened 2015-05-26
[DNM] General project cleanup

lolletsoc:
@kbairak @glezos


kbairak:
@lolletsoc is this ready?

lolletsoc:
No, not yet

lolletsoc:
@kbairak Review and merge please! We'll need to do another PR to deal with documentation as well.

openformats/utils/__init__.py
```diff
@@ -0,0 +1,36 @@
+from ..exceptions import RuleError
+
+_RULES_ATOI = {
+    'zero': 0,
+    'one': 1,
+    'two': 2,
+    'few': 3,
+    'many': 4,
+    'other': 5
+}
```
kbairak: @lolletsoc could it be that some format had different names fo these? Will we be able to override them if that's the case? I don't know how likely this scenario is, but this was the reason I put this as a Handler attribute

lolletsoc: Hmm, fair enough. I'll make a change now.

kbairak:
nice, merging

PR #21, opened 2015-05-26
Testbed preselect sample source

kbairak:
This has a bit of javascript, but I think it would be good to be merged, especially for future format implementations.


kbairak:
@glezos @lolletsoc one of you two press the button when ready

PR #24, opened 2015-05-29
Docs getting started

kbairak:
The main difference is that I moved the documentation into the docstrings and had this page fetch those instead.


kbairak:
@lolletsoc I think it's ready

PR #30, opened 2015-09-10
Take 'product' attribute into account

kbairak:
Also, make the compiler work and provide the automatic tests


kbairak:
@amanatidis Hey, remember this? :D

This has to do with [this issue](https://transifex.atlassian.net/browse/TX-3976).

Any chance you can review this? I'm about to copy paste a big part of it to txc again to resolve the ticket.

Thanks!

kbairak:
I don't understand why the tests fail. It has something to do with installing dependencies. I'll look into it when I find some time. Locally they pass though.

PR #33, opened 2016-03-21
Json

kbairak:
TODOs:

- [x] Decide whether to use `DumbJson` or the official `json` module
- [x] Clean up some leftover code
- [x] Escape keys
- [x] Return helpful `ParseError`s when things go wrong
- [x] Use official `json` module to do a first pass
- [x] check for JSON values that aren't only arrays, dicts or strings (make sure booleans, nulls, dates etc are handled properly)
- [x] See if we have to take care of case where root JSON element is not dict
- [x] Write tests

@tabac @ollandos @SofiaMargariti @kouk I need your help in deciding the first point

pros of `DumbJson`:
- It allows us to preserve indentation and raw (escaped) source strings

cons of `DumbJson`:
- It's handwritten, not official.

@tabac @ollandos If you're up for an early review (although there's still work left to do), feel free.

### DIFF
--- a/bin/create_files.py
+++ b/bin/create_files.py
@@ -7,11 +7,15 @@ Example:
     $ ./bin/create_files.py openformats/tests/srt/files/1_en.srt
 """
 
+from __future__ import absolute_import
+
 import argparse
 import os
 import sys
+
 sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
-from openformats.formats import (plaintext, srt, android)
+
+from openformats.formats import plaintext, srt, android, json
 from openformats.tests.utils import translate_stringset
 
 args = argparse.ArgumentParser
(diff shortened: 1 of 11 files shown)


kbairak:
I'm going with the `DumbJson` option until given compelling reason not to.

kouk:
@kbairak +1 for dumbjson. Although tbh I would prefer if it was separated from the actual handler code. Like a separate library project.

kbairak:
@kouk:

The coverage falling makes sense considering there's a new handler with no tests. I will write them.

It is a separate library, it is on `utils/json.py`

kouk:
@kbairak right you are. I got confused with the `JsonHandler_` code still being in there.

kbairak:
I think this is done. Anybody wants to review?

PR #34, opened 2016-04-05
Escape/Unescape support.

tabac:


openformats/formats/android.py
```diff
@@ -345,3 +345,17 @@ def _compile_plurals(self, plurals_tag, plurals_offset):
             # didn't find it, must remove by skipping it
             self.transcriber.skip_until(plurals_offset +
                                         len(plurals_tag.content))
+
+    @staticmethod
+    def escape(string):
+        escaped_string = string.replace('"', r'\"'). \
+                                replace("'", r"\'")
```
kbairak: do we need to worry about single quotes too?

kbairak:
```python
In [37]: print json.dumps("hello'world")
"hello'world"
```

kbairak: ah, we were in android, not json, sorry

kbairak:
For JSON, I think we should only stick to double-quotes and backslashes.

As for Android, I think more thought is required...

I guess that this source:

```xml
<resources>
  <string name="foo">click &gt;a href="..."&lt;here&gt;/a&lt;</string>
</resources>
```

Should appear like this:

```
click <a href="...">here</a>
```

in raw mode.

I guess we should do a few tests with the official android handler to see what it does.

kbairak:
Here are my findings on how the official Android handler handles un/escaping:

```
In [51]: c.put('/project/android_escape/resource/res1/content/', json={'content': '<resources><string name="foo">click<a href="here">here</a></string></resources>'})
Out[51]: <Response [200]>
In [52]: c.get('/project/android_escape/resource/res1/translation/en/strings/').json()
Out[52]:
[{u'comment': u'',
  u'context': u'',
  u'key': u'foo',
  u'pluralized': False,
  u'reviewed': False,
  u'source_string': u'click<a href="here">here</a>',
  u'translation': u'click<a href="here">here</a>'}]
In [53]: c.put('/project/android_escape/resource/res1/content/', json={'content': '<resources><string name="foo"><a></b></string></resources>'})
Out[53]: <Response [400]>
In [54]: print _53.text
Could not import file: Opening and ending tag mismatch: a line 1 and b, line 1, column 38
```

```
In [55]: c.put('/project/android_escape/resource/res1/content/', json={'content': '<resources><string name="foo">&gt;a&lt;</string></resources>'})
Out[55]: <Response [200]>
In [56]: c.get('/project/android_escape/resource/res1/translation/en/strings/').json()
Out[56]:
[{u'comment': u'',
  u'context': u'',
  u'key': u'foo',
  u'pluralized': False,
  u'reviewed': False,
  u'source_string': u'&gt;a&lt;',
  u'translation': u'&gt;a&lt;'}]
```

```
In [57]: from lxml import etree
In [58]: s = 'click &lt;a href="here"&gt;here&lt;/a&gt;'
In [59]: etree.fromstring('<a>{}</a>'.format(s)).text
Out[59]: 'click <a href="here">here</a>'
```

tabac:
@kbairak take another look.

openformats/formats/json.py
```diff
@@ -248,3 +248,15 @@ def _get_next_string(self):
             return self.stringset[self.stringset_index]
         except IndexError:
             return None
+
+    @staticmethod
+    def escape(string):
+        escaped_string = string.replace('\\', r'\\').replace('"', '\\"')
```
kbairak: Not terribly important, but for consistency's sake I would stick with `r''` notation everywhere.

openformats/formats/json.py
```diff
@@ -248,3 +248,15 @@ def _get_next_string(self):
...
+
+        return escaped_string
+
+    @staticmethod
+    def unescape(string):
+        unescaped_string = string.replace(r'\\', '\\').replace(r'\"', '"')
```
kbairak: same here

kbairak:
Looks good. I would add a few tests for un/escaping backslashes and I would stick to `r''` notation everywhere.

Should I merge or do you want to give these a try?

tabac:
@kbairak so what you have in mind is this?

```python
    @staticmethod
    def escape(string):
        escaped_string = string.replace('\\', r'\\').replace(r'"', r'\"')

        return escaped_string

    @staticmethod
    def unescape(string):
        unescaped_string = string.replace(r'\\', '\\').replace(r'\"', r'"')

        return unescaped_string
```

About the tests, I have added one for each case, more?

PR #35, opened 2016-04-07
Json whitespace

kbairak:
@tabac how about we give this a go before we tag master?

### DIFF
--- a/openformats/utils/json.py
+++ b/openformats/utils/json.py
@@ -57,7 +57,8 @@ class DumbJson(object):
     def __init__(self, source, start=0):
         self.source = source
         self._end = None
-        starting_symbol, self.start = self._find_next('{[', start)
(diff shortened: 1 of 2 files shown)


openformats/utils/json.py
```diff
@@ -162,14 +167,16 @@ def _iter_list(self):
...
+                ',]', next_p, require_whitespace=True,
+            )
             if next_symbol == ',':
                 start = next_symbol_p + 1
             elif next_symbol == ']':
                 self.end = next_symbol_p
                 break
 
-    def _find_next(self, symbols, start=0):
+    def _find_next(self, symbols, start=0, require_whitespace=False):
```
tabac: Why not make if by default `True`

```python 
def _find_next(self, symbols, start=0, require_whitespace=True):
```

that way we won't have to write it everywhere.

PR #37, opened 2016-04-15
DNM: Dumbxml2

kbairak:
Opening up for comments.

### DIFF
--- a/openformats/utils/xml.py
+++ b/openformats/utils/xml.py
@@ -240,6 +240,11 @@ class NewDumbXml(object):
 
     @property
     def position(self):
+        """The starting position of the tag.
+
+        <atag>Some text</atag>
+        ^
+        """
         if self._position is not self.NOT_CACHED:
             return self._position
 
@@ -248,6 +253,11 @@ class NewDumbXml(object):
 
     @property
     def tag(self):
+        """The name of the tag.
+
+        <atag>Some text</atag>
+         ^^^^
+        """
         if self._tag is not self.NOT_CACHED:
             return self._tag
 
@@ -299,6 +309,11 @@ class NewDumbXml(object):
 
     @property
     def text_position(self):
+        """The start position of the text.
+
+        <atag>Some text</atag>
+              ^
+        """
         if self._text_position is not self.NOT_CACHED:
             return self._text_position
 
@@ -377,6 +392,11 @@ class NewDumbXml(object):
 
     @property
     def tail_position(self):
+        """The position just after the tag.
+
+        <atag>Some text</atag>|
+                              ^
+        """
         if self._tail_position is not self.NOT_CACHED:
             return self._tail_position
 
@@ -392,6 +412,11 @@ class NewDumbXml(object):
 
     @property
     def tail(self):
+        """The text that follows the tag untill the start of a news one.
+
+        <atag>Some text</atag> newlines etc <anothertag>...
+                              ^^^^^^^^^^^^^^
+        """
         if self._tail is not self.NOT_CACHED:
             return self._tail
 
@@ -401,6 +426,11 @@ class NewDumbXml(object):
 
     @property
     def end(self):
+        """The starting position of the next tag.
+
+        <atag>Some text</atag> newlines etc <anothertag>...
+                                            ^
+        """
         return self.tail_position + len(self.tail)
 
     def find_children(self, *tags):
--- a/openformats/utils/xml.py
+++ b/openformats/utils/xml.py
@@ -369,7 +369,9 @@ class NewDumbXml(object):
                 raise ValueError("Invalid closing of tag '{}'".
                                  format(self.tag))
             else:
-                inner = NewDumbXml(self.source, start)
+                # Use `self.__class__` in case this is a subclass (eg to handle
+                # HTML)
+                inner = self.__class__(self.source, start)
                 yield inner
                 start = inner.end
 
--- a/openformats/utils/xml.py
+++ b/openformats/utils/xml.py
@@ -243,7 +243,7 @@ class NewDumbXml(object):
         if self._position is not self.NOT_CACHED:
             return self._position
 
-        self._position = self._find_next_gt(self.start)
+        self._position = self._find_next_lt(self.start)
         return self._position
 
     @property
@@ -338,7 +338,7 @@ class NewDumbXml(object):
             self._text = None
             return self._text
 
-        next_tag_position = self._find_next_gt(self.text_position)
+        next_tag_position = self._find_next_lt(self.text_position)
         if next_tag_position == len(self.source):
             raise ValueError("Tag '{}' not closed".format(self.tag))
 
@@ -396,7 +396,7 @@ class NewDumbXml(object):
             return self._tail
 
         self._tail = self.source[self.tail_position:
-                                 self._find_next_gt(self.tail_position)]
+                                 self._find_next_lt(self.tail_position)]
         return self.tail
 
     @property
@@ -415,7 +415,7 @@ class NewDumbXml(object):
             for inner in child.find_descendants(*tags):
                 yield inner
 
-    def _find_next_gt(self, start):
+    def _find_next_lt(self, start):
         in_cdata = False
         for ptr in xrange(start, len(self.source)):
             candidate = self.source[ptr]
--- a/openformats/utils/xml.py
+++ b/openformats/utils/xml.py
@@ -401,26 +401,16 @@ class NewDumbXml(object):
     def end(self):
         return self.tail_position + len(self.tail)
 
-    def find_children(self, tag_or_tags=None):
-        if isinstance(tag_or_tags, (str, unicode)):
-            tags = [tag_or_tags]
-        elif tag_or_tags is not None:
-            tags = tag_or_tags
-
+    def find_children(self, *tags):
         for child in self:
-            if tag_or_tags is None or child.tag in tags:
+            if not tags or child.tag in tags:
                 yield child
 
-    def find_descendants(self, tag_or_tags=None):
-        if isinstance(tag_or_tags, (str, unicode)):
-            tags = [tag_or_tags]
-        elif tag_or_tags is not None:
-            tags = tag_or_tags
-
+    def find_descendants(self, *tags):
         for child in self:
-            if tag_or_tags is None or child.tag in tags:
+            if not tags or child.tag in tags:
                 yield child
-            for inner in child.find_descendants(tag_or_tags):
+            for inner in child.find_descendants(*tags):
                 yield inner
 
     def _find_next_gt(self, start):
--- a/openformats/utils/xml.py
+++ b/openformats/utils/xml.py
@@ -402,6 +402,8 @@ class NewDumbXml(object):
     def content_end(self):
         if self._content_end is not self.NOT_CACHED:
             return self._content_end
+        if self.text_position is None:
+            return None
         for _ in self:
             pass
         return self._content_end
@@ -410,6 +412,8 @@ class NewDumbXml(object):
     def content(self):
         if self.tag == self.COMMENT:
             return self.text
+        if self.content_end is None:
+            return None
         return self.source[self.text_position:self.content_end]
 
     @property
--- a/openformats/utils/xml.py
+++ b/openformats/utils/xml.py
@@ -189,6 +189,10 @@ class NewDumbXml(object):
         - `dumb_xml.text_position`: The starting point of the tag's contents
           (in our example: `15`)
 
+            - For single tags (eg '<br />'), `text` and `text_position` will be
+              None; for empty tags (eg, '<li></li>'), `text` will be `''` and
+              `text_position` will be the expected value
+
         - `dumb_xml.text`: The text content of the tag (in our example: `'b'`)
 
             - If the contents begin with text but also have other xml tags, we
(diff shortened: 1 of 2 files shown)


kbairak:
@ollandos: Styilng-related: I think that all these constants and some of the helper methods sacrifice readability and compactness (if that's a word) and do not offer anything in return in terms of structure.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+        parsed = NewDumbXml(source)
+        children_itterator = self._get_children_itterator(
+            parsed,
+            *[
+                self.STRING,
+                self.STRING_ARRAY,
+                self.STRING_PLURAL,
+                NewDumbXml.COMMENT
+            ]
+        )
```
kbairak: 1. the `*` doesn't offer anything, the above is equivalent to `self._get_children_iterator(parsed, self.STRING, self.STRING_ARRAY, self.STRING_PLURAL, NewDumbXml.COMMENT)`.

2. As I said in the conversation, I would prefer if we simply called, `parsed,find_children()`

3. As I said in the conversation, I believe the `STRING`, `STRING_ARRAY` and `PLURALS` constants don't offer anything compared to using the words themselves.

kbairak: You might come back with "hey, *you* are doing it for comments"; sure, but `'!--'` is confusing to someone who reads this code months from now, I think that `'string-array'` will not be.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
-            content = content.decode("utf-8")  # convert to unicode
-
-        resources_tag_position = content.index("<resources")
+        resources_tag_position = content.index(self.PARSE_START)
 
         self.transcriber = Transcriber(content[resources_tag_position:])
-        source = self.transcriber.source
+        stringset = []
+        self.current_comment = None
+        self.order = 1
```
kbairak: 1. I don't know if it's a "standard" or anything, but I think we usually start counting from `0` for the order

2. I think it's better to use `itertools.count` for this (I did it for the json format). Try this:

```python
import itertools
c = itertools.count()
print c.next()  # => 0
print c.next()  # => 1
print c.next()  # => 2
```

openformats/formats/android.py
ollandos: The rule of thumb is usually use is that if a literal is used more than once in your code just declare it. If someday you want to change it for some reason you will be thankful you did it that way. But let's discuss it further on Monday.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
-            # didn't find it, must remove by skipping it
-            self.transcriber.skip_until(plurals_offset +
-                                        len(plurals_tag.content))
+        :param child: The child to retrieve the attributes from.
+        :returns: Returns a tuple (name, product)
+        :raises: It raises a parse error if no `name` attribute is present.
+        """
+        name = child.attrib.get('name')
+        if name is None:
+            raise ParseError("")
```
kbairak: Very minor: I think a `try/catch` would be more descriptive of what's going on

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+        product = child.attrib.get('product')
```
kbairak: IIRC, the `context` attribute should either be an empty or a non-empty string. I *think* some things go very wrong when the strings go into transifex for weird reasons. Try `...attrib.get('product', "")`

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
-            self.transcriber.skip(len(tag.inner))
-        else:
-            self.transcriber.copy_until(offset + tag.inner_offset +
-                                        len(tag.inner))
+            self.transcriber.skip(len(child.text))
 
-        # ...ello World</string>
-        #                       ^
-        self.transcriber.copy_until(offset + len(tag.content))
+        return [string]
```
kbairak: Later on you do a `self.order += len(strings)`. This `len()` will return 1, even if `string is None`. I think you should default to returning `[]`

kbairak: probably won't matter if you end up using `itertools.count` for the order

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+            return self._handle_string_array(child)
+        elif child.tag == self.STRING_PLURAL:
+            return self._handle_string_plural(child)
+        elif child.tag == NewDumbXml.COMMENT:
+            self._handle_comment(child)
+        return None
+
+    def _handle_string(self, child):
+        name, product = self._get_child_attributes(child)
+        string = self._create_string(
```
kbairak: I don't think you need a constructor method to create Openstrings, Openstring has a constructor.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+            name,
+            child.text,
```
kbairak: This is interesting. If the string tag is like this:

```xml
<string name="foo">hello <b>world</b></string>
```

`.text` will return only `"hello "`. I think I need to implement another attribute to `DumbXml` called `.content` or something.

kbairak: I just did it, `.content` should be fine here. (I should write tests for it too though)

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+
+        children_itterator = self._get_children_itterator(
+            child, *[self.STRING_ITEM]
+        )
+        string_rules_text = {}
+        for new_child in children_itterator:
+            rule = new_child.attrib.get('quantity')
+            if rule is None:
+                raise ParseError("")
+            rule_number = self.get_rule_number(rule)
```
kbairak: This might raise a "`RuleError`" if the quantity is not recognized. Perhaps we should catch that and convert it to a `ParseError`.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+            string_rules_text[rule_number] = new_child.text
```
kbairak: `.content`

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+            product,
+            plural=True
+        )
+        if string is not None:
             # <plurals name="foo">   <item>Hello ...
-            #                        ^
-            self.transcriber.copy_until(plurals_offset + first_item_offset)
-
+            #
+            first_plural_position = child.text_position + len(child.text)
```
kbairak: You could have stored it during the iteration above like:

```python
first_plural_position = None
for new_child in ...:
    first_plural_position = first_plural_position or child.position
    # ...
```

That's fine as it is however, no real preference.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+            self.transcriber.copy_until(first_plural_position)
+            self.transcriber.add(string.template_replacement)
             # ...</item>   </plurals>...
             #           ^
-            self.transcriber.add(string.template_replacement)
-            self.transcriber.skip(last_item_offset +
-                                  len(last_item_tag.content) -
-                                  first_item_offset)
+            self.transcriber.skip(
+                new_child.tail_position - first_plural_position
```
kbairak: 1. `new_child` is the last iterated tag from before, right? That sounds a bit scary

2. You can use `self.transcriber.skip_until(new_child.tail_position)` to get the same effect.

openformats/formats/android.py
```diff
@@ -1,349 +1,228 @@
...
+            )
 
-        else:
-            string = None
+        return [string]
```
kbairak: Same as before, `len(strings)` will be `1`, even if `string is None`

kbairak:
@ollandos amazing job!!! Most of my comments are in terms of code-styling. Since it's all subjective, you can ignore those. But we can talk about them.

openformats/formats/android.py
kbairak: 3 conter-arguments, in case I forget them on Monday:

1. since the constants' names are the same as their values, if Google changes the file-format and we come back here to change our code, my guess is that we will change the names of the constants too

2. My experience/opinion says that if you're writing DRY code, the "more than once" is too aggressive and leads to more abstractions than necessary and the code ends up more complicated and unreadable. I think that "more than twice" or even "3 times" is much better. Also see [this](http://c2.com/cgi/wiki?YouArentGonnaNeedIt).

3. This chilren-iteration only happens once.

openformats/formats/android.py
ollandos: Obviously I used 0 when I started but the UI returns Order: none in the string so I thought I was wrong.

openformats/formats/android.py
ollandos: True I missed it.

openformats/formats/android.py
ollandos: Already changed it to rule error.

openformats/formats/android.py
ollandos: Yeah for some reason I skipped that property of the transcriber.

openformats/formats/android.py
ollandos: Needs to be content here as well right?

openformats/formats/android.py
ollandos: It is the last iterated child. skip_util is another method I missed apparently.

openformats/formats/android.py
ollandos: Ack. Will fix.

openformats/formats/android.py
ollandos: I am doing some empty strings check in the _create_string method. But I'm thinking of moving the checks to a _validate_not_empty(string, plural=False) method. If you agree on this will remove the _create_string completely.

openformats/formats/android.py
ollandos: And I just read that content was not implemented when I was working sooooo. But yes a content method sounds awesome.

ollandos: Also I feel the text method has a bit misleading name as it only presents the text before the inner tags. In cases like these:
text1<a></a>text2<a></a>text3
text will only return the first one.

openformats/formats/android.py
ollandos: I thought the opposite :D. But no biggie, will do.

openformats/formats/android.py
ollandos: 1. True (If the one fixing it is not lazy)
2. Reading...
3. We use them also to compare in the _handle_string method.

openformats/formats/android.py
ollandos: itertools.count() doesn't allow skipping values :(. And I do not want to keep the counter in every function. Prefer it in one clear place.

openformats/utils/xml.py
```diff
@@ -143,3 +143,388 @@ def _is_within_comment(self, match):
...
+    @property
+    def position(self):
+        """ The starting position of the tag.
+
+            <atag>Some text</atag>
+            ^
+        """
+
+        if self._position is not self.NOT_CACHED:
+            return self._position
```
tabac: @kbairak we have this almost in every property, maybe we should do what you did with `core.permissions`, something like a `cached_property` decorator.

openformats/utils/xml.py
```diff
@@ -143,3 +143,388 @@ def _is_within_comment(self, match):
...
+            <atag>Some text</atag>
+             ^^^^
+        """
+
+        if self._tag is not self.NOT_CACHED:
+            return self._tag
+
+        if self.source[self.position:
+                       self.position + len("<!--")] == "<!--":
+            self._tag = self.COMMENT
```
tabac: Although I would prefer

```python
self.source[self.position:].startswith("<!--")
```

apparently what you have here is faster!

openformats/utils/xml.py
```diff
@@ -143,3 +143,388 @@ def _is_within_comment(self, match):
...
+            self._process_comment()
+            return self._tag
+
+        for ptr in xrange(self.position + 1, len(self.source)):
+            candidate = self.source[ptr]
+            if candidate in ('/', '>') or candidate.isspace():
+                self._tag = self.source[self.position + 1:ptr]
+                return self._tag
+        raise ValueError("Opening tag not closed")
```
tabac: I think we should have our own Exceptions here too, just like in the handlers.

openformats/utils/xml.py
kbairak: seriously?

I will change it to yours; readability is its own reward

kbairak: I think I get it. Yours copies a bit chunk of the source string to a different place. Now I'm sceptical...

kbairak: I'll keep mine, I'll assign start and end to variables to make the end result a bit prettier

openformats/utils/xml.py
kbairak: IIRC, I raised the same exception that `lxml` uses. I'll check again

kbairak: Nope, I was wrong. Fixing.


```
In [12]: etree.fromstring('<a')
  File "<string>", line unknown
XMLSyntaxError: Couldn't find end of Start Tag a line 1, line 1, column 3
```

openformats/utils/xml.py
kbairak: I tried playing around with this, but we sometimes need to control the order of how these properties are calculated and this decorator would mess things up a bit. I'd advise against it

kbairak:
@tabac did 2/3 of your suggestions (and tried to do the third). I'd say it's all yours to merge.

PR #38, opened 2016-04-19
Ignore empty strings in KEYVALUEJSON format

kbairak:
@ollandos spare two minutes to review n' merge this? :heart_eyes_cat:


openformats/formats/json.py
```diff
@@ -71,6 +73,8 @@ def _extract(self, parsed, nest=None):
                 else:
                     key = u"{}..{}..".format(nest, index)
                 if isinstance(item, (str, unicode)):
```
ollandos: a small thing here. I know it is not relevant to the current PR but this:
isinstance(value, (str, unicode))

can be like this:
isinstance(value, basestring)

ollandos:
Also again on a general note (not related to the PR).
Line 55-68 and 73-87 look extremely similar? Why not make a method out of them?

Anyway besides those comments everything looks legit to me.

PR #39, opened 2016-04-20
Android v2

kbairak:
Opening for comments


openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+        :param child: The child to retrieve the attributes from.
+        :returns: Returns a tuple (`name`, `product`)
+        :raises: It raises a ParseError if no `name` attribute is present.
+        """
+        name = child.attrib.get('name')
+        if name is None:
+            raise ParseError(
+                'Missing the `name` attribute on line'.format(
+                    self.transcriber.line_number
+                )
```
kbairak: The transcriber has counted the number of newlines it has copied/skipped up to this point. We don't know when the last copy/skip operation took place. Perhaps a `self.transcriber.copy_until(child.position)` would be in order before calling `line_number`

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+        if isinstance(text, dict):
+            if len(text) == 0:
+                return False
+            # If there is plural missing raise error.
+            for key, string in text.iteritems():
+                if string.strip() == "":
+                    raise ParseError(
+                        'Missing plural string before the line {}'.format(
+                            self.transcriber.line_number
+                        )
```
kbairak: Same comment about `line_number`. Also, I think we can be a bit more descriptive with the error message, something like `Missing form for plural string '{name}', quantity '{quantity}' on line {line_number}`.

kbairak: You will probably need to pass the whole plural tag instead of the text here in order to achieve these.

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+        """Creates a string and returns it. If empty string it returns None.
+
+        :param text: The strings text.
+        :param name: The name of the string.
+        :param comment: The developer's comment the string might have.
+        :param product: Extra context for the string.
+        :returns: Returns an OpenString object if the text is not empty
+                  else None.
+        """
+        if self._validate_not_empty(text):
```
kbairak: We also have to make sure no duplicate strings are found. The `<name, product>` pair must be unique or things will go wrong along the road. Keep a `self.existing_somethings` set somewhere and check against that. You can use the Openstrings' `.string_hash` property to ensure uniqueness (or `.template_replacement` but as you'll see, they're almost identical)

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+        children_itterator = child.find_children(self.STRING_ITEM)
+        # Itterate through the children with the item tag.
+        for new_child in children_itterator:
+            rule = new_child.attrib.get('quantity')
+            if rule is None:
+                # If quantity is missing the plural is unknown
+                raise ParseError(
+                    "Missing the `quantity` attribute on line {}".format(
+                        self.transcriber.line_number
+                    )
```
kbairak: Same comment about `line_number`

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
                 )
-                self._order += 1
-                yield string
-
-            # ... <item>Hello...
+            rule_number = self.get_rule_number(rule)
```
kbairak: I guess we should catch the `RuleError` here and raise a ParseError instead.

kbairak: Come to think of it, I don't remember if we do this in the JsonHandler ... :fearful:

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+            # <plurals name="foo">   <item>Hello ...
+            #                        ^
+            first_plural_position = child.text_position + len(child.text)
+            self.transcriber.copy_until(first_plural_position)
+            self.transcriber.add(string.template_replacement)
+            # ...</item>   </plurals>...
             #           ^
-            self.transcriber.copy_until(string_array_offset + item_offset +
-                                        item_tag.inner_offset)
+            self.transcriber.skip_until(new_child.tail_position)
```
kbairak: I guess if someone uploads something like `<plurals></plurals>`, new_child will never get a value and you'll get an error here, how about `.skip_until(child.content_end)`?

openformats/formats/android.py
kbairak: oh wait, JSON doesn't have plurals :relieved:

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+
+        :returns: An list containing the OpenString objects if any were created
+                    else None.
+        """
+        strings = []
+        children_itterator = child.find_children(self.STRING_ITEM)
+        name, product = self._get_child_attributes(child)
+        # Itterate through the children with the item tag.
+        for index, new_child in enumerate(children_itterator):
+            child_name = u"{}[{}]".format(name, index)
```
kbairak: ok, I just had a scary thought. btw this wasn't take care of the previous version of beta-android either. What if you have a source file like this:

```xml
<resources>
  <string-array name="a"><item>hello</item></string-array>
  <string name="a[0]">goodbye</string>
</resources>
```

Should we escape the `[` into the `name` attributes into `\[`?

openformats/formats/android.py
ollandos: If the the plurals tag is empty the _create_string() function will return None and it won't reach this line.
Check line 229.

kbairak:
@ollandos: Another comment, code-styling-related, as optional as ever:

The `child` and `new_child` variable names can get a bit confusing IMO. I would rather they were something like `string_tag` or `plurals_node` or `item_xml` or something (obviously with consistent suffixes).

openformats/formats/android.py
kbairak: gotcha!

ollandos:
I agree.

openformats/formats/android.py
ollandos: Why would we escape it? Can you elaborate more on this?

kbairak: In the above example, both strings will get the `a[0]` key. Even if you skip checking for key uniqueness, when you try to save this to the database or upload a language file or compile, things will get messy.

kbairak: If you escape the names hoever, "hello" will get `a[0]` and "goodbye" will get `a\[0]`

kbairak: You'll need to escape the backslashes too ... :/

ollandos: Got it. Will do.

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+            return
+
+        # Check if any string matches array items
+        not_empty = False
+        for new_child in children_itterator:
+            if self._should_compile(new_child):
+                not_empty = True
+                break
+
+        if not_empty:
```
kbairak: Very minor: the double-negative is kinda confusing, how about using an `empty` variable instead?

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
-    def _compile_plurals(self, plurals_tag, plurals_offset):
+            # Remove the `string-array` tag
+            self._skip_tag(child)
+
+    def _compile_string_plural(self, child):
+        """
+        """
+        # Check if child was empty to begin with
+        if child.content.strip() == '':
+            return
```
kbairak: If a `<plurals>` was empty to begin with, parsing should not have even completed; is this necessary?

openformats/formats/android.py
```diff
@@ -1,349 +1,431 @@
...
+            splited_content = child.content.split(
+                self.next_string.template_replacement
+            )
+            start = splited_content[0]
+            end = splited_content[1]
+
+            # If newline formating
+            if start.startswith(end):
+                start = start.replace(end, '', 1)
+                self.transcriber.add(end)
```
kbairak: Something very clever must be happening here (not sarcasm). I will try to mess around with it to figure things out, but maybe a few more comments would be in order?

kbairak:
@ollandos: Everything's beautiful! Perhaps it's time to drop the old ones and have the test-suite use this one. One final suggestion: how about the parser checks whether the `.text` and `.tail` of `<string-array>`s and `<plurals>` (and their items) is empty or raise an error? I can parse something like this:

```xml
<resources>
    <string-array name="a">
        hello <item>a</item> world
    </string-array>
</resources>
```

ollandos:
Not sure what should be allowed. If the android specifications do not allow it I think we should raise an error.

openformats/formats/android.py
ollandos: Will do.

openformats/formats/android.py
ollandos: Right now I allow empty plurals tag in the parsing. Should I not?

openformats/formats/android.py
ollandos: Hahaha. Alright.

ollandos: Ooooor maybe ```not_not_empty```?

openformats/formats/android.py
kbairak: Right, you just don't collect it. I guess it's right. Sorry.

kbairak:
Regardless of whether it's allowed, it messes things up for plurals. It doesn't explicitly say in the documentation that it is **not** allowed, but I think it's safe to assume that it's not.

[Here]() is the documentation, it will come in handy when implementing the `escape` and `unescape` methods.

openformats/formats/android.py
kbairak: then, line 345 would become `if not not_not_empty:`

openformats/formats/android.py
kbairak: good thing you have a centralized method for retrieving the `name` and `product` :smile:

kbairak:
Also, put the link to the documentation in the class's docstring

openformats/formats/android.py
```diff
@@ -3,13 +3,21 @@
 import itertools
 
 from ..handlers import Handler
-from ..exceptions import ParseError
-from openformats.strings import OpenString
-from openformats.transcribers import Transcriber
+from ..strings import OpenString
 from ..utils.xml import NewDumbXml
+from ..exceptions import RuleError
+from ..exceptions import ParseError
```
kbairak: these can go in the same line, I guess

openformats/formats/android.py
```diff
@@ -212,14 +237,36 @@ def _create_string(self, name, text, comment, product):
                 order=self.order_counter.next(),
                 developer_comment=comment
             )
+            # If duplicate hash raise ParseError
+            if string.string_hash in self.existing_hashes:
+                self.transcriber.copy_until(child.position)
+                if not product:
+                    msg = (
+                        u"Duplicate `name` ({}) property found on the "
+                        "line {}. Specify a the `product` to differentiate"
```
kbairak: concaternate unicode strings, if `line_number` had a non-ascii character (I know it won't but it's the principle), this would raise an error

openformats/formats/android.py
```diff
@@ -212,14 +237,36 @@ def _create_string(self, name, text, comment, product):
...
+                    ).format(
+                        name, self.transcriber.line_number
+                    )
+                else:
+                    msg = (
+                        u"Duplicate `name` ({}) and `product` ({}) "
+                        "properties found on the line {}"
```
kbairak: same

openformats/formats/android.py
```diff
@@ -231,9 +278,18 @@ def _validate_not_empty(self, text):
...
+                    self.transcriber.copy_until(child.tail_position)
+                    last_line = self.transcriber.line_number
                     raise ParseError(
-                        'Missing plural string before the line {}'.format(
-                            self.transcriber.line_number
+                        u'Missing plural string between lines {}-{}, '
+                        'on item with `quantity`: {}'.format(
+                            first_line,
+                            last_line,
+                            self.get_rule_string(key)
```
kbairak: maybe consider the case where `first_line` is the same as `last_line`?

PR #40, opened 2016-05-09
Android v2 escape

kbairak:
Added un/escape functions to Android format

According to rules described here:
http://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling

Check this out:

```
In [13]: c.create_new_resource('android', 'android', 'BETA_ANDROID', content=u'''
             <resources>
                <string name="backslashed">a\\'b</string>
                <string name="doublequoted">"a'b"</string>
            </resources>
         ''')
Out[13]: <Response [201]>

In [14]: # After translating both strings with `α'β` in "rich" mode on transifex:

In [14]: print c.get('/project/android/resource/android/translation/el/').json()['content']

<resources>
    <string name="backslashed">α\'β</string>
    <string name="doublequoted">α\'β</string>
</resources>
```

### DIFF
--- a/openformats/formats/android.py
+++ b/openformats/formats/android.py
@@ -21,7 +21,7 @@ class AndroidHandler(Handler):
     name = "ANDROID"
     extension = "xml"
 
-    EXTRACTS_RAW = False
+    EXTRACTS_RAW = True
 
     # Where to start parsing the file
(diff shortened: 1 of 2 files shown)


kbairak:
@tabac you're up!

PR #41, opened 2016-05-10
Be sure that child.text is not None

tabac:
@kbairak plz review, merge.

### DIFF
--- a/openformats/formats/android.py
+++ b/openformats/formats/android.py
@@ -419,7 +419,7 @@ class AndroidHandler(Handler):
             elif child.tag == self.STRING_PLURAL:
                 self._compile_string_plural(child)
 
-    def _compile_string(self, child, string_array=False):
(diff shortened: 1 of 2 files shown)


kbairak:
There's the same issue with the `.content` property:

```
In [1]: from openformats.formats.android import AndroidHandler

In [2]: h = AndroidHandler()

In [3]: h.parse('<resources><string name="a" /></resources>')

...

/Users/kbairak/devel/repos/transifex/txformats/openformats/formats/android.py in _validate_not_empty(self, text, child)
    324                 )
    325                 self._raise_error(child, msg)
--> 326         elif text.strip() == "":
    327             return False
    328         return True

AttributeError: 'NoneType' object has no attribute 'strip'
```

kbairak:
@tabac ^^

tabac:
Closing we'll open a PR against devel branch.

PR #49, opened 2016-05-17
Parsing/Compiling/Some testing for the stringsdict format.

kbairak:
Opening for comments


openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,389 @@
...
+            self.transcriber.skip_until(value_tag.end)
+
+            self.transcriber.add(
+                key_tag.tail +
+                self.PLACEHOLDER_KEY +
+                key_tag.tail +
+                self.PLACEHOLDER_STRING.format(
+                    string.template_replacement
+                )
+                + value_tag.tail
```
kbairak: Minor: in python it's better to `"".join((a, b, c))` than to `a + b + c`. I don't remember the article I read it in, I think it has to do with the fact that python knows how much space to allocate for the final string.

openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,389 @@
...
+    KEY = "key"
+
+    """ Parse Methods """
+
+    def parse(self, content):
+        self.transcriber = Transcriber(content)
+        self.order_counter = itertools.count()
+        source = self.transcriber.source
+        # Skip xml info declaration
+        resources_tag_position = content.index(self.PARSE_START)
```
kbairak: force of habit :) I guess you mean `dict_tag_position`

openformats/utils/xmlutils.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def reraise_syntax_as_parse_errors(func):
+    def safe_parse(*args, **kwargs):
+        try:
+            template, stringset = func(*args, **kwargs)
+        except DumbXmlSyntaxError as e:
+            raise ParseError(unicode(e))
+        return template, stringset
+    return safe_parse
```
kbairak: Maybe make this a bit more generic? like

```python
def reraise_as_parse_errors(exc_type):
    @functools.wraps(func)
    def decorator(func):
        def decorated(*args, **kwargs):
            try:
                return_value = func(*args, **kwargs)
            except exc_type, e:
                raise ParseError(unicode(e))
            return return_value
        return decorated
    return decorator
```

and use it like:

```python
@reraise_as_parse_error(DumbXmlSyntaxError)
def parse(self, content):
   # ...
```

kbairak: Oh, another thing. What about prepending "XML syntax error: " to the error message?

openformats/utils/xmlutils.py
```diff
@@ -0,0 +1,154 @@
+from ..handlers import Handler
+from ..exceptions import ParseError
+from ..utils.xml import NewDumbXml, DumbXmlSyntaxError
+
+
+def reraise_syntax_as_parse_errors(func):
```
kbairak: use `@functools.wraps(func)` here. I don't know much about it, but it's a good practice

openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,427 @@
...
+
+    """ Parse Methods """
+
+    @reraise_syntax_as_parse_errors
+    def parse(self, content):
+        self.transcriber = Transcriber(content)
+        self.order_counter = itertools.count()
+        source = self.transcriber.source
+        # Skip xml info declaration
+        dict_tag_position = source.index(self.PARSE_START)
```
kbairak: This is why I don't like overusing class-level variables. I was confused for quite some time because I thought this referred to the `<plist` tag. I had to go to another place to figure it out. I'm not proposing a change btw, just expressing my thoughts.

openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,427 @@
...
+                            the key uniqueness.
+        :raises: ParseError if the tag is not a <key>. Also raises a ParseError
+                    if main_key is True and the key is not unique.
+        :returns: A string containing the <key> tags content.
+        """
+        if key_tag.tag != self.KEY:
+            message = (
+                u"Was expecting <key> tag but found <{tag}> tag "
+                u"on line {line_number}"
+            )
```
kbairak: Shouldn't `key` be a formatting variable too? 😛

openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,427 @@
...
+                    created the list is empty.
+        """
+        # The first key tag contains the main key
+        main_key = self._handle_key(key_tag, main_key=True)
+        dict_iterator = self._handle_dict(dict_tag)
+
+        string_list = []
+        for key_child in dict_iterator:
+            # The second key contains the secondary key.
+            secondary_key = self._handle_key(key_child)
```
kbairak: I would not leave a default value to the `main_key` argument and specify it every time.

openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,427 @@
...
+
+    """ Compile Methods """
+
+    def compile(self, template, stringset):
+        dict_tag_position = template.index(self.PARSE_START)
+
+        self.transcriber = Transcriber(template[dict_tag_position:])
+        source = self.transcriber.source
+
+        parsed = NewDumbXml(source)
```
kbairak: How come we're not doing the

```python
self.transcriber = Transcriber(template)
template = self.transcriber.template
dict_tag_position = template.index(self.PARSE_START)
parsed = NewDumbXml(template, dict_tag_position)
```

trick here?

openformats/formats/stringsdict.py
```diff
@@ -0,0 +1,427 @@
...
+        self.transcriber.copy_until(dict_tag.end)
+
+    def _compile_plural_string(self, placeholder_key, placeholder_value):
+        """Replaces the placeholder tags with the translation strings.
+
+        :param placeholder_key: The placeholder key tag.
+        :param placeholder_value: The placeholder value tag.
+        :NOTE: Assigns the self property `next_string` to the next OpenString.
+        """
+        string_list = self.next_string.string.items()
```
kbairak: Shouldn't these be sorted? It seems that for dicts with integer keys, `items()` returns them sorted, at least for values 0 to 5 with which we're dealing with here, but if feels more right to sort them anyway.

kbairak:
I'm done with my comments. Looks good in general.

openformats/formats/stringsdict.py
ollandos: Not sure what you mean.

openformats/formats/stringsdict.py
ollandos: Doing the same with android. Not sure why I went to do it like this. this time. Will change it to be more consistent.

openformats/formats/stringsdict.py
ollandos: You are right. Will sort.

openformats/formats/stringsdict.py
ollandos: No reason (if you are not kidding). Checking for key here so I can just hardcode it :P

openformats/utils/xmlutils.py
ollandos: Will read a bit on it first and will apply.

PR #50, opened 2016-05-19
Added validation of int in transcriber to avoid shenanigans

ollandos:


openformats/transcribers.py
```diff
@@ -268,3 +272,10 @@ def edit_newlines(self, chunk, enforce_newline_type=None):
             return chunk.replace('\n', '\r\n')
         else:
             return chunk
+
+    def _validate_int_type(integer):
```
kbairak: you forgot `self`, or you forgot to make this a `@staticmethod`

PR #53, opened 2016-05-27
Po

ollandos:


openformats/formats/po.py
```diff
@@ -0,0 +1,161 @@
...
+                string = entry.msgid if pluralized else {
+                    '0': entry.msgid,
+                    '1': entry.msgid_plural
+                }
+            else:
+                raise ParseError(u"")
+        elif not self.only_keys:
+            self.only_values = True
+        else:
+            raise ParseError(u"")
```
kbairak: the silent treatment ... 😄

kbairak:
I can't find something wrong here (the test suite can though)

openformats/formats/po.py
ollandos: Not sure what to write there as an Error message :P

kbairak: "You have both full and empty `msgstr`s, make up your mind already"

openformats/formats/po.py
```diff
@@ -0,0 +1,238 @@
...
+    EXTRACTS_RAW = False
+
+    def parse(self, content, is_source=False):
+        stringset = []
+        self.is_source = is_source
+        self.order_generator = itertools.count()
+        po = polib.pofile(content)
+        self.only_values = False
+        self.only_keys = False
+        self.new_po = copy.copy(po)
```
kbairak: Why the copy?

openformats/formats/po.py
```diff
@@ -0,0 +1,238 @@
...
+        :returns: The string of the entry.
+        :raises: ParseError if incosistency is found on the file. To be more
+                    verbose: Either all msgstr attributes are fille or none is.
+        """
+        if pluralized:
+            string = entry.msgstr_plural
+        else:
+            string = entry.msgstr
+
+        is_empty = self._validate_empty(entry, string, pluralized)
```
kbairak: @ollandos Very minor: If it's a method that raises an error if there's trouble otherwise does nothing, I call it `_validate_foo`. If it's a method that returns True or False, I call it `_is_foo`.

kbairak: scratch that, it does both... sorry

kbairak:
Looks good, I vote merge!

openformats/formats/po.py
ollandos: I use the copy to remove the entries. If I didn't and removed an item while looping it would get fd up.

kbairak:
@ollandos Latest changes look great!

PR #60, opened 2016-06-10
Small fixes

ollandos:


openformats/formats/po.py
```diff
@@ -179,7 +179,9 @@ def _get_string(self, entry, pluralized):
                     verbose: Either all msgstr attributes are fille or none is.
         """
         if pluralized:
-            string = entry.msgstr_plural
+            string = {}
+            for k, v in entry.msgstr_plural.iteritems():
+                string[int(k)] = v
```
tabac: You could use a dict comprehension:

```python
string = {int(k): v for k, v in entry.msgstr_plural.iteritems()}
```

openformats/strings.py
```diff
@@ -92,7 +92,12 @@ def __hash__(self):
         return hash((self.key, self.context, self.rule))
 
     def __repr__(self):
-        return '"{}"'.format(self._strings[5].encode('utf-8'))  # pragma: nocover # noqa
+        return next(
+            (
+                self._strings.get(i) for i in xrange(5, -1, -1)
+                if self._strings.get(i)
+            )
+        ).encode('utf-8')
```
tabac: I think we shouldn't assume that there's always going to be at least one value in `self._strings`. I feel the version below is more straightforward.

```python
for i in xrange(5, -1, -1):
    try:
        return self._strings[i].encode('utf-8')
    except KeyError:
        pass

return 'something went wrong'
```

PR #63, opened 2016-11-17
android plural fix for same tag_names

kbairak:
@kbairak cloud you plz review it? 

Description:

Assume we have the following source content: 
```<?xml version="1.0" encoding="utf-8"?>
<resources>
  <string name="test"> this is just a test string</string>
  <plurals name="test">
    <item quantity="one"> One</item>
    <item quantity="other"> other</item>
  </plurals>
</resources>
```
If we try to compile it the Compiler gives -> `Duplicate "name" (test) attribute found on line 4. Specify a "product" to differentiate `

Since it's a valid XML we should handle it and allow it as source content.

### DIFF
--- a/openformats/formats/android.py
+++ b/openformats/formats/android.py
@@ -70,7 +70,7 @@ class AndroidHandler(Handler):
             NewDumbXml.COMMENT
         )
         stringset = []
-        self.existing_hashes = set()
+        self.existing_hashes = {}
         for child in children_itterator:
             strings = self._handle_child(child)
             if strings is not None:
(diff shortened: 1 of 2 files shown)


ChrsMark:
@kbairak this is the different approach we discussed, and seems to do the job. [Ignore failing tests since I will not fix the until we finalize we want it like this].

openformats/formats/android.py
```diff
@@ -245,6 +245,26 @@ def _create_string(self, name, text, comment, product, child,
...
+                    format_dict = {
+                        'name': name,
+                        'product': product,
+                        'child_tag': child.tag
+                    }
+                    msg = (
+                        u"Duplicate `tag_name` ({child_tag}) for `name`"
+                        u" ({name}) and `product` ({product}) "
+                        u"found on line {line_number}"
+                    )
```
kbairak: product could be empty, maybe a small `if` to style the message differently?

ChrsMark: Fixed

kbairak:
looks good, merging. Lets deploy this on monday with a new version of `master`.

PR #64, opened 2016-12-14
Android compile non translatable

kbairak:
[TX-7091](https://transifex.atlassian.net/browse/TX-7091)

When compiling android files for translation, the handler should skip the tags marked with `translatable="false"`.

This adds a new kwarg to the compile method of **all** handlers, `is_source`. This is what tells the handler to either include or skip the untranslatable strings.

After this is merged, we should make a [change to transifex](https://github.com/transifex/txc/pull/3892) to pass this option when invoking the compile method.

### DIFF
--- a/openformats/formats/android.py
+++ b/openformats/formats/android.py
@@ -351,7 +351,7 @@ class AndroidHandler(Handler):
 
     """ Compile Methods """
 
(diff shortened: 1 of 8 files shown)


ChrsMark:
LGTM merge

PR #65, opened 2016-12-21
Better handling of backslashes on JSON

kbairak:
Previously, in strings like: `{"a": "b\\"}`, the parser would consider
the last double-quote as escaped, and thus not know when the string
ends. In fact, the first backslash escapes the second one and the
double-quote should be considered as the end of the string.

This fix makes sure the first backslash escapes the character after it
and, if the character after it is a backslash too, the second one does
not escape the next character.

### DIFF
--- a/openformats/tests/util_tests/test_json.py
+++ b/openformats/tests/util_tests/test_json.py
@@ -72,6 +72,22 @@ class DumbJsonTestCase(unittest.TestCase):
         self._test_dfs('["a", {"b": "c"}]',
                        [("a", 2), ([("b", 8, "c", 13)], 6)])
 
+    # Escaping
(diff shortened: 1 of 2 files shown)


kbairak:
Hey @ollandos . I think you're the one who's the most somewhat familiar with what's going on here. can you take a look?

Thelma needs this for a customer tomorrow, if you can't review it, let me know or pass it to someone else.

kbairak:
oh, [related helpscout ticket](https://secure.helpscout.net/conversation/294276134/25978/?folderId=243634)

kbairak:
and related [jira ticket](https://transifex.atlassian.net/browse/TX-7277)

ollandos:
Smart. Merging!

PR #68, opened 2017-03-02
Github markdown file format support

diegobz:


kbairak:
@diegobz try putting that sample file in the proper place in order to auto-generate some tests and have the testbed pre-load it.

http://openformats.readthedocs.io/en/latest/testing.html#get-yourself-a-sample-file

kbairak:
@diegobz nice, btw. I'll check it out

diegobz:
@kbairak Done!

kbairak:
I will merge this, but please make sure that, after the demo, we delete the resources created and work thoroughly on the format before putting it into production. Because few people understand it and there are not sufficient tests, I wouldn't say it's production ready and I wouldn't want to introduce second versions that early.

PR #70, opened 2017-05-08
GitHub markdown continue

SofiaMargariti:
* Parse template cell into separate strings
* Maintain indentation for code blocks indented with 4 or more spaces
* Improve YAML parser

### DIFF
--- a/openformats/formats/github_markdown.py
+++ b/openformats/formats/github_markdown.py
@@ -97,9 +97,11 @@ class TxBlockLexer(BlockLexer):
                     table_token = self.tokens[-1]
                     keys = table_token.keys()
                     if 'header' in keys and 'cells' in keys:
-                        self.md_stringset.extend([(h, 'header') for h in table_token['header']])
                         self.md_stringset.extend(
-                            [(cell, 'cell') for row in table_token['cells'] for cell in row]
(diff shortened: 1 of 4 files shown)


openformats/formats/github_markdown.py
```diff
@@ -8,12 +8,25 @@
 from ..utils.compilers import OrderedCompilerMixin
 
 
-def string_handler(string):
+def string_handler(token, template):
```
kroustou: I think a doc string is needed here.

SofiaMargariti: yes, in general need some better documentation. I will fix that

openformats/formats/github_markdown.py
```diff
@@ -79,9 +111,20 @@ def manipulate(text):
                 # methods to avoid getting duplicated parts of the markdown
                 # content in the `self.md_stringset` because of the recursion.
                 parser_rules = ('list_block', 'def_footnotes')
-                if key and key not in parser_rules:
+                table_rules = ('table', 'nptable')
```
kbairak: Minor: Since these two are constant, they could be defined on the top of the file instead of being re-assigned everytime inside the for-loop

PR #72, opened 2017-08-21
upgrade mistune to next version and use expandtabs

SofiaMargariti:
Mistune expands tabs to spaces and removes white characters from blank lines, so we need to do the same so that we are able to match the strings in the original content.


SofiaMargariti:
I understand that but in that case the only alternative would be to check the original file and if it contained tabs to not accept it and through an error. Do we want to be so strict?

SofiaMargariti:
@kbairak we've talked with @diegobz about this and we think it is ok for the tabs to be expanded to spaces. Can we merge this?

PR #73, opened 2017-08-25
TX-8481 KEYVALUEJSON plurals

dbendilas:
https://transifex.atlassian.net/browse/TX-8481

Added support for plurals as per ICU's message format (plural subset).

e.g.

```
{
  "singular_key": "This is a regular string.",
  "total_files": "{ item_count, plural, one {You have {file_count} file.} other {You have {file_count} files.} }",
  "gold_coins": "{ count, plural, zero {The chest is empty.} one {You have one gold coin.} other {You have {cnt} gold coins.} }"
}
```

## Unsupported features

The following ICU features are *not* supported:
- the `offset` feature 
- the explicit count rule, e.g. `=0`, `=1`


openformats/formats/json_plurals.py
```diff
@@ -0,0 +1,506 @@
...
+      - the `offset` feature
+      - the explicit count rule, e.g. `=0`, `=1`
+    """
+
+    name = "KEYVALUEJSON_PLURALS"
+    extension = "json"
+
+    MESSAGE_FORMAT_STRUCTURE = re.compile(
+        ur'{\s*([A-Za-z-_\d]+),\s*([A-Za-z_]+)\s*,\s*(.*)}'
+    )
```
kbairak: {minor, opinionated] When something is not used often, I don't like using a constant for it. It makes me have to search for it when I encounter it later on.

openformats/formats/json_plurals.py
```diff
@@ -0,0 +1,506 @@
...
+    the following features are *not* supported:
+      - the `offset` feature
+      - the explicit count rule, e.g. `=0`, `=1`
+    """
+
+    name = "KEYVALUEJSON_PLURALS"
+    extension = "json"
+
+    MESSAGE_FORMAT_STRUCTURE = re.compile(
+        ur'{\s*([A-Za-z-_\d]+),\s*([A-Za-z_]+)\s*,\s*(.*)}'
```
kbairak: shouldn't the regex start and end with `^`, `$` to include the whole string, and also take possible spaces at the start/end of the string? I added a space character at the start of a ICU-plural string in the testbed and it treated the whole string as a regular string

kbairak: Instead of that dot (`.*`) at the end of the regex, I would prefer a `[^}]*` (anything but a closing bracket, zero or more times)

openformats/formats/json_plurals.py
```diff
@@ -0,0 +1,506 @@
...
+        # Make sure the plurality rules are valid
+        # If not, an error will be raised
+        if len(valid_matches) != len(all_matches):
+            self._handle_invalid_plural_format(
+                serialized_strings, any_plural_item, key, value
+            )
+
+        # Create a list of tuples [(plurality_str, content_with_braces)]
+        all_strings_list = [
+            JsonPluralsHandler._parse_plural_content(match[0])
```
kbairak: why not `self`?

openformats/formats/json_plurals.py
```diff
@@ -0,0 +1,506 @@
...
+            for match in valid_matches
+        ]
+
+        # Convert it to a dict like { 'one': '{...}', 'other': '{...}' }
+        # And then to a dict like { 1: '...', 5: '...' }
+        all_strings_dict = dict(all_strings_list)
+        all_strings_dict = {
+            Handler.get_rule_number(plurality_str): content[1:-1]
```
kbairak: why not `self`?

openformats/formats/json_plurals.py
```diff
@@ -0,0 +1,506 @@
...
+            for plurality_str, content in all_strings_dict.iteritems()
+        }
+
+        openstring = OpenString(key, all_strings_dict, order=next(self._order))
```
kbairak: set `pluralized=True` in the kwargs; some languages have only one plural rule so a `{ item_count, plural, other { only this one here } }` string is still valid ICU-plural

openformats/formats/json_plurals.py
```diff
@@ -0,0 +1,506 @@
...
+        return at_least_one
+
+    def _insert_plural_string(self, value, value_position, string,
+                              is_real_stringset):
+        templ_replacement = string.template_replacement
+        replacement_pos = value.find(templ_replacement)
+
+        if is_real_stringset:
+            replacement = \
+                JsonPluralsHandler.serialize_pluralized_string(
```
kbairak: why not `self`?

kbairak:
I have to brush my pyparsing skills before I review the rest. However, this brings me to my question: isn't pyparsing overkill? From what I remember, "regular" languages can be processed by regular expressions and "the other kind" (context-free or something) languages need something like pyparsing to be processed. I think regular expressions are sufficient to process ICU-plurals. I will investigate and return for more.

openformats/formats/json_plurals.py
dbendilas: Somehow I missed that. I added `\s*` both at the beginning and the end and it works fine. I don't think `^` and `$` are necessary, and they didn't do anything on their own.

openformats/formats/json_plurals.py
dbendilas: I agree with you. I thought I would add a lot of similar consts along the way, but it turned out I only needed this one. I'm using the string directly now

openformats/formats/json_plurals.py
dbendilas: Fixed.

openformats/formats/json_plurals.py
dbendilas: Fixed.

openformats/formats/json_plurals.py
dbendilas: Didn't now that, thanks. Fixed.

openformats/formats/json_plurals.py
dbendilas: Fixed.

dbendilas:
> I have to brush my pyparsing skills before I review the rest. However, this brings me to my question: isn't pyparsing overkill? From what I remember, "regular" languages can be processed by regular expressions and "the other kind" (context-free or something) languages need something like pyparsing to be processed. I think regular expressions are sufficient to process ICU-plurals. I will investigate and return for more.

I tried matching the format using regular expressions, but found it pretty hard. I researched it quit a bit and all of the sources I found said that it's impossible or next to impossible. The issue I was having had to do with matching multiple substrings that look like this `{ <optional { }> <optional { }> }`

Many of the answers I found online suggested pyparsing, and in the end that's what I used. I'd be very interested to see how the same thing can be accomplished using regular expressions. That said, it's probably not worth changing now, to be honest. I wouldn't expect any dramatic change in performance.

dbendilas:
> Simply enhance the old handler to handle ICU-plurals.

Yes, I've done that now. @kbairak can you please test yourself a few cases to make sure there is no backwards incompatibility? I'd really appreciate an extra set of eyes on this, to be sure I haven't missed anything.

kbairak:
I did some tests with regular expressions and come to the same conclusion as you.

Here are a few cases I've tested in the testbed:

```
// 1. Original
{"a": "{a, plural, one { {a} file } other { {a} files }}"}

// 2. Space at the beginning of the string => Captured as singular string
{"a": " {a, plural, one { {a} file } other { {a} files }}"}

// 3. Space after the first '{' => Fine
{"a": "{ a, plural, one { {a} file } other { {a} files }}"}

// 4. Space before the first ',' => Captured as singular string
{"a": "{a , plural, one { {a} file } other { {a} files }}"}

// 5. Space before the second ',' => Fine
{"a": "{a, plural , one { {a} file } other { {a} files }}"}

// 6. Space before the last '}' => Fine
{"a": "{a, plural, one { {a} file } other { {a} files } }"}

// 7. Space at the end of the string => Fine
{"a": "{a, plural, one { {a} file } other { {a} files }} "}

// 8. Two spaces before 'one' => Fine
{"a": "{a, plural,  one { {a} file } other { {a} files }}"}

// 9. Two spaces after 'one' => Fine, but during compilation the two spaces become one (probably can't be helped)
{"a": "{a, plural, one  { {a} file } other { {a} files }}"}

// 10. Two spaces before 'other' => Fine, but during compilation the two spaces become one (probably can't be helped)
{"a": "{a, plural, one { {a} file }  other { {a} files }}"}

// 11. Two spaces after 'other' => Fine, but during compilation the two spaces become one (probably can't be helped)
{"a": "{a, plural, one { {a} file } other  { {a} files }}"}
```

I believe we don't have to worry about the latter ones, they can't be helped. But I think we can accommodate `2` and `4`.

dbendilas:
> I believe we don't have to worry about the latter ones, they can't be helped. But I think we can accommodate 2 and 4.

I fixed both of these cases. I was supposed to have fixed case 2 yesterday, but I missed something in the regex. I added test cases for both.

@kbairak

diegobz:
@dbendilas @kbairak do we also support `selectordinal` besides `plural`?

https://formatjs.io/guides/message-syntax/#selectordinal-format
http://icu-project.org/apiref/icu4j/com/ibm/icu/text/MessageFormat.html

dbendilas:
> do we also support selectordinal besides plural?

@diegobz no, we don't. Should we? I didn't realise this was in the specs.

diegobz:
@dbendilas If it's not too much of a hassle, I think it would be a good idea.

PR #74, opened 2017-08-30
Version 2 of GithubMarkdownHandler using an actual yaml parser

SofiaMargariti:
Jira: https://transifex.atlassian.net/browse/TX-8511
Implement yaml parser based on https://github.com/transifex/txc/blob/devel/transifex/resources/formats/yml/yml.py
and use this parser for the frontmatter header of the GFM files.

screen shot 2017-08-31 at 14 11 56


openformats/tests/formats/github_markdown_v2/files/1_tpl.md
```diff
@@ -0,0 +1,156 @@
...
+  - list_key:
+    - ac56af108b4b69b867c09f8542a51cb2_tr
+    - d2d20cce249c5d7e8ca96b9e30a3349c_tr
+  - 17e2086737d2192e6665ea2c0f0b1e9a_tr
+key2:
+  - object_within_list: 1a397fe7999e432441bfce7a1f4fb1f3_tr
+  - aa8c6113a4236d576f96ba9b0bd4dfa4_tr
+  - 76d7c5ccc4301f10b39115b2c1a9ca47_tr
+
+description: fdba4efe4ab9d38afee4054a7fa06e29_trcustom_vars:
```
diegobz: @SofiaMargariti looks like we have a missing `\n` here.

SofiaMargariti: @diegobz yes, this is known behaviour. The solution I was thinking is to append a new line during compilation in this type of strings, but another way is to lower the end index of this node by one during parsing, so the new line is considered part of the template.

PR #76, opened 2017-09-12
Wrap problematic translations in quotes

SofiaMargariti:
Github markdown header is parsed a yaml document. Characters `:`, `&`, `*` and `-` have special meaning in a yaml file so the translations that contain them should be wrapped in double quotes.


SofiaMargariti:
@dbendilas good catch! yes we should, I will fix that

SofiaMargariti:
@dbendilas it is fixed

PR #80, opened 2017-10-03
Fix markdown translation file compiling error.

fathineos:
Will affect both markdown v1 & v2. The problem that solves is the mismatch
of hashes (template replacement) during parsing step and compiling step, due
to inconsistency of handling.


nelefth:
https://transifex.atlassian.net/browse/TX-8470

PR #81, opened 2017-10-09
Handle 'tools:locale' attribute of 'resources' tag if present

kbairak:
Related to [TX-8565: The value of tools attribute should indicate the locale of the file](https://transifex.atlassian.net/browse/TX-8565)

# Checklist (for the reviewer)

* [x] Code is well documented
* [ ] [does not apply] Performs well when applied to big organizations/resources/etc from our production database

# Problem

Android specs define a `tools:locale` attribute at the top `<resources>` tag of the file which should be set to the language code currently used in the application. In our case, this means that when compiling, we should set this attribute to the code of the language we're currently compiling against.

# Steps to reproduce

1. Select/create a project with the English source language
2. Upload the following file using the Android file format
   ```xml
   <resources tools:locale="en">
     <string name="hello">world</string>
   </resources>
   ```
3. Translate
4. Download the translation file; the `tools:locale` attribute of the downloaded file will still be `"en"` instead of the code of the target language

# Solution

The custom XML library we implemented for the Android file format (and others), takes care to not only process the content of the files, but also the positions of the elements. However, this approach did not extend to the attributes of the tags; instead the `.attrib` property simply returned a dictionary of the attributes, like in `lxml`.

The first commit of this PR changes how tag attributes are handled. We introduce an `.attributes` property which returns a list of attributes where each attribute is represented as a `(key_position, key, value_position, value)` tuple. We keep the `.attrib` property too for backwards compatibility.

The second commit, uses this new functionality to replace the `tools:locale` attribute during compilation, if such an attribute is found. The transcriber is used for this, as usual.

NOTE: A branch is required on TXC to make sure that the `language_info` kwarg is passed when compiling (https://github.com/transifex/txc/pull/4553)

# Example run / Screenshots

(Not needed)

# Performance on live data

(Not needed)


fathineos:
@kbairak Would it make sense to add attributes in the testbed?

openformats/formats/android.py
```diff
@@ -355,13 +355,26 @@ def _get_child_attributes(self, child):
...
+                **kwargs):
         resources_tag_position = template.index(self.PARSE_START)
 
         self.transcriber = Transcriber(template[resources_tag_position:])
         source = self.transcriber.source
 
         parsed = DumbXml(source)
+
+        # Check against 'tools:locale' attribute
+        if language_info is not None and 'tools:locale' in parsed.attrib:
```
fathineos: Maybe we should handle the case of empty string ""

kbairak: You mean as an attribute or if `language_info['code'] == ""`?

In the first case, I think the best thing for us to do is to fill the attribute nevertheless.

In the second case, I assume that if someone took care to pass `language_info` it should be well-formed

kbairak:
> @kbairak Would it make sense to add attributes in the testbed?

I don't know, would it? In any case, I think we're fine for now. It's been a while since I worked on the testbed, it would take some time to re-familiarize myself with it.

knordbo:
I believe the ideal solution is to not copy over `tools:locale` to the translated files, AFAIK `tools:locale` is used to define the default locale. See: https://plus.google.com/+TorNorbye/posts/NJTDSqy77We and https://developer.android.com/studio/write/tool-attributes.html#toolslocale

This attribute is not needed in the translations. In android lint checks can infer the translated languages since it follows a predictable locale package structure like `values-fr/strings.xml`, but lint doesn't know you base language since it lives under `values/strings.xml`.

TLDR; The best solution is to ignore `tools:locale` in any translations. I'm happy to make the small change if you agree.

PR #84, opened 2017-11-14
Skips replacement for fences inside list_blocks

igavriil:
Problem
-------
Compilation for markdown that includes fences (code blocks) inside list blocks fails.
This is due to inconsistency created on the parsing step, where the template replacement
is failing to match extracted fences from the template, with those that actually exist 
in the template.

Steps to reproduce
------------------
Parse and compile the following md template:
<pre>
1. List
   1. ```
      code
      ```
</pre>

Solution
--------
Skips template replacement when extracted substring cannot be found on template.

### DIFF
--- a/openformats/formats/github_markdown.py
+++ b/openformats/formats/github_markdown.py
@@ -215,7 +215,7 @@ class GithubMarkdownHandler(OrderedCompilerMixin, Handler):
         curr_pos = 0
         for string in (yaml_stringset + block.md_stringset):
             string = string_handler(string, template)
-            if string:
+            if string and string in template[curr_pos:]:
                 string_object = OpenString(str(order), string, order=order)
                 order += 1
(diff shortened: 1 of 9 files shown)


openformats/formats/github_markdown.py
```diff
@@ -215,7 +215,7 @@ def parse(self, content, **kwargs):
         curr_pos = 0
         for string in (yaml_stringset + block.md_stringset):
             string = string_handler(string, template)
-            if string:
+            if string and string in template[curr_pos:]:
```
SofiaMargariti: @igavriil There is another version of the github markdown parser in `github_markdown_v2.py` (the two versions only change in the way they parse the yaml header). Can you apply the same change [there](https://github.com/transifex/openformats/blob/devel/openformats/formats/github_markdown_v2.py#L120) as well?

PR #85, opened 2018-01-10
TX-9113 Indesign format

SofiaMargariti:
Related to TX-9113

Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------
We need to add support for Adobe's Indesign file format. Indesign files are exported to an IDML format [1].

Steps to reproduce
------------------

Solution
--------
The .idml file is unpacked using UCFlib [2]. All translatable text is located inside the story xml fragments (everything that starts with a key `Stories/Story_<id>.xml`). The Stories content is parsed and the strings are extracted and replaced with their replacement hash. Finally the fragments are packed back into their initial format and in order for the resource template to be created.

Example run / Screenshots
-------------------------
Here is a sample xml content of a Story fragment:
```xml
<?xml version="1.0" encoding="utf-8"?>
<idPkg:Story DOMVersion="13.0" xmlns:idPkg="http://ns.adobe.com/AdobeInDesign/idml/1.0/packaging">
    <Story AppliedNamedGrid="n" AppliedTOCStyle="n" IsEndnoteStory="false" Self="u2f25" StoryTitle="$ID/" TrackChanges="false" UserText="true">
        <StoryPreference FrameType="TextFrameType" OpticalMarginAlignment="false" OpticalMarginSize="12" StoryDirection="LeftToRightDirection" StoryOrientation="Horizontal"/>
        <InCopyExportOption IncludeAllResources="false" IncludeGraphicProxies="true"/>
        <ParagraphStyleRange AppliedParagraphStyle="ParagraphStyle/body_sans">
            <CharacterStyleRange AppliedCharacterStyle="CharacterStyle/$ID/[No character style]" FillColor="Color/C=0 M=87 Y=3 K=0" FontStyle="Regular">
                <Properties>
                    <AppliedFont type="string">Midnight Chalker</AppliedFont>
                </Properties>
                <Content>+ <?ACE 7?></Content>
            </CharacterStyleRange>
            <CharacterStyleRange AppliedCharacterStyle="CharacterStyle/$ID/[No character style]">
                <Content>Tell the truth</Content>
                <Br/>
            </CharacterStyleRange>
        </ParagraphStyleRange>
    </Story>
</idPkg:Story>
```
Everything that is enclosed in `<Content>..</Content>` tags is considered a string. So from the above xml there would be extracted two string:
* `+ <?ACE 7?>`
* `Tell the truth`

[1] http://wwwimages.adobe.com/www.adobe.com/content/dam/acom/en/devnet/indesign/sdk/cs6/idml/idml-specification.pdf
[2] https://github.com/davidwtbuxton/ucflib


openformats/formats/indesign.py
```diff
@@ -119,6 +124,30 @@ def _can_skip_content(self, string):
             pass
         return False
 
+    def _preserve_symbols(self, xml):
```
SofiaMargariti: @rigaspapas  Even with this solution we still had issues because it resulted to `&amp;amp;#x200A;` in the final string. If you check my next commits, I decided to simplify a little bit the parse method. Now we don't use both regular expressions and Beautiful soup but only a simple regular expression that finds all the strings inside the <Content> tags and replaces them in place with the correct template replacement and returns the replaced xml content.

openformats/tests/formats/indesign/test_idml.py
```diff
@@ -0,0 +1,38 @@
+import unittest
+
+from openformats.formats.indesign import InDesignHandler
+
+
+class InDesignTestCase(unittest.TestCase):
```
dbendilas: I think that we need more unit tests of some of the internal functionality of the handler. Also edge cases like special characters etc.

openformats/tests/formats/indesign/files/sample.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dbendilas: What is the purpose of this file? Is it used anywhere? Is it there to show what the binary `sample.idml` file contains? If this is the text content of the idml sample, then I would say it's not a very good sample for testing.

SofiaMargariti: This was used on a test that I removed. I will update the tests as you ask in the next comment and if I don't find a use for this file, I will remove it.

requirements.txt
```diff
@@ -3,3 +3,8 @@ polib==1.0.3
 mistune==0.7.3
 PyYAML==3.10
 pyparsing==2.2.0
+
+# InDesign
+lxml==4.1.1
```
kbairak: I didn't find any use for `lxml` in this branch. Even if there was, I would object with using both `lxml` and `BeautifulSoup` at the same time. Even now, I prefer `lxml` over `BeautifulSoup`, but this is arguably a personal opinion. However, I think `lxml` is faster.

setup.py
```diff
@@ -7,6 +7,9 @@
     'mistune==0.7.3',
     'PyYAML==3.10',
     'pyparsing==2.2.0',
+    'lxml==4.1.1',
```
kbairak: See above

openformats/formats/indesign.py
```diff
@@ -0,0 +1,176 @@
...
+        replaces it with the string_hash and returns the resulting
+        template while updating `self.stringset` in the process.
+        args:
+            story_xml (str): The xml content of a single Story of the IDML file
+        returns:
+            the input string with all translatable content replaced by the
+            md5 hash of the string.
+        """
+
+        template = self.CONTENT_REGEX.sub(self._replace, story_xml)
```
kbairak: I remember [this](https://stackoverflow.com/a/1732454/1511263). In short it says that it's always a bad idea to parse XML with regular expressions. I tend to agree, although in openformats we have problems with xml parsers too because we want to capture the entirety of a tag's content, regardless of whether it contains other tags or not. That's why DumbXml was created.

For example, if the file has a `<Content>` tag inside a `<Content>` tag, this implementation will run into trouble.

So, question: Is there a reason we are using this instead of DumbXml with transcriber? A sample implementation would be:

```python
transcriber = Transcriber(source)
source = transcriber.source
start = content.index('<idPkg')
parsed = DumbXml(content, start)
for content_tag in parsed.find_descendants('Content'):
    string = content_tag.content
    openstring = OpenString(string, string, next(self.order))
    self.stringset.append(openstring)
    
    transcriber.copy_until(content_tag.text_position)
    transcriber.skip(len(string))
    transcriber.add(openstring.template_replacement)

transcriber.copy_until(len(source))
```

openformats/formats/indesign.py
```diff
@@ -0,0 +1,176 @@
...
+        """ Implements the logic used by `self.CONTENT_REGEX.sub(...)` to
+        replace strings with their template replacement and appends new strings
+        to `self.stringset`.
+        """
+        opening_tag = match.group(1)
+        string = match.group(2).decode('utf-8')
+        closing_tag = match.group(3)
+        if self._can_skip_content(string):
+            return opening_tag + string + closing_tag
+        string_object = OpenString(string, string, order=self.order)
```
kbairak: I think that using the string as the key will create problems during translation file uploads and downloads. Have we tried this with txc? The key should be something that is identical in a source file and a language file. I think a good candidate for this would be the XPATH of the Content tag. I know that lxml can provide that:

```python
In [15]: a = etree.fromstring('<a>hello <b>world</b></a>')


In [16]: a.getroottree().getpath(a[0])
Out[16]: '/a/b'
```

openformats/formats/indesign.py
```diff
@@ -0,0 +1,176 @@
...
+        translations_dict = {s.template_replacement: s for s in stringset}
+        hash_regex = re.compile('[a-z,0-9]{32}_tr')
+
+        # Iterate over the contents of the IDML file
+        for key in idml.keys():
+            if not key.startswith("Stories/"):
+                continue
+
+            story_content = idml[key]
+            for match in hash_regex.finditer(story_content):
```
kbairak: If the order of the xml files within the archive is the same as during parsing, and the order of the strings in the stringset is the same as during parsing, then we can simply search for the first hash (using python), replace it, search for the second hash and so on. See the `UnorderedCompilerMixin` for an example.

openformats/formats/indesign.py
```diff
@@ -0,0 +1,176 @@
...
+        """ Parses .idml file content and returns the resource template and
+            stringset.
+            * Use UCF to unpack `content` to xml fragments
+            * Parse all Story fragments to extract the translatable strings
+              and replace them with a replacement hash
+            * Pack the fragments back to create the template
+            * Return the (template, stringset) tuple
+        """
+        self.stringset = []
+        self.order = 0
```
kbairak: One more minor thing: We have found that it's better to use `self.order = count()` (`from itertools import count`). And then apply the order like `OpenString(..., next(self.order))`. True, this will only save you one line of code (the `+= 1` part), but I believe it looks better and expresses what's going on well enough.

openformats/formats/indesign.py
```diff
@@ -0,0 +1,176 @@
...
+    def _can_skip_content(self, string):
+        """
+        Checks if the contents of an XML files are translateable.
+        Strings that contain only special characters or can be evaluated
+        to a nunber are skipped.
+        """
+        if not self.SPECIAL_CHARACTERS_REGEX.sub('', string).strip():
+            return True
+        try:
+            float(string)
```
kbairak: How about `string.strip()`?

openformats/formats/indesign.py
SofiaMargariti: The reason I went with this approach was to avoid escaping / unescaping the special characters ([see here](https://mayart.de/download/Indesign-IDML/special-idml-chars.pdf)). For `<?ACE 7?>` for example which is parsed as a separate element I would have to somehow concatenate it with the rest of the content and for special characters like `&#x200A;` I couldn't avoid it being turned into their equivalent unicode character `\u200a`.

In the idml specification is says in the description of the `<Content>` tag:
> The content of the `<CharacterStyleRange>`. Can be a string of text or a reference to an inline element.

References are expressed in the attributes of an XML tag in IDML, so I don't think the content tag can have any other sub-elements.

I tried the above example with NewDumbXml but it also has issues with the <?ACE .. ?> characters 
```
/Users/Sofia/environments/txc/lib/python2.7/site-packages/openformats-0.0.33-py2.7.egg/openformats/utils/xml.pyc in __iter__(self)
    463                         u"Closing tag '{}' does not match opening tag '{}' on "
    464                         u"line {}".
--> 465                         format(closing_tag, self.tag, self._find_line_number())
    466                     )
    467                 for ptr in xrange(start + 2 + len(self.tag), len(self.source)):

DumbXmlSyntaxError: Closing tag 'Char' does not match opening tag '?ACE' on line 80
```

kbairak: yeah, ok. perhaps DumbXml could be "fixed" in a backwards-compatible way but it's probably not worth it. According to the specs then, the regex approach is fine. And since these files are not written by hand by anyone, but rather by software, we should be safe. Although, trusting users to supply correct files has proven troublesome in the past 😄

openformats/formats/indesign.py
SofiaMargariti: @kbairak I ended up using the order of the string, since Content tags between different stories can have the same xpath and it would require to parse the xml a second time just to get the path.

_Each story has a unique ID inside the IDML document but this ID is not preserved between import / export in InDesign so it would be different in the translation file)._

openformats/formats/indesign.py
SofiaMargariti: I updated the compile method to handle the strings in the same order it is done in `parse`.

requirements.txt
SofiaMargariti: I think lxml was need because we use BeautifulSoup with the `xml` option, but still don't need both. I've removed beautifulsoup and kept lxml.

SofiaMargariti:
@kbairak @dbentilas pls take another look at the latest changes.

openformats/formats/indesign.py
```diff
@@ -0,0 +1,181 @@
...
+        return template
+
+    def _replace(self, match):
+        """ Implements the logic used by `self.CONTENT_REGEX.sub(...)` to
+        replace strings with their template replacement and appends new strings
+        to `self.stringset`.
+        """
+        opening_tag = match.group(1)
+        string = match.group(2).decode('utf-8')
+        closing_tag = match.group(3)
```
kbairak: Very-very-very minor: `opening_tag, string, closing_tag = match.groups()`

openformats/formats/indesign.py
```diff
@@ -0,0 +1,181 @@
...
+
+        if self._can_skip_content(string):
+            return opening_tag + string + closing_tag
```
kbairak: Very-very-very minor: `return match.group()`

openformats/formats/indesign.py
```diff
@@ -0,0 +1,181 @@
...
+        order = next(self.order)
+        string_object = OpenString(str(order), string, order=order)
+        self.stringset.append(string_object)
+        return opening_tag + string_object.template_replacement + closing_tag
```
kbairak: Very-very-very minor: I *think* that `"".join((a, b, c))` is faster than `a + b + c`. Something to do with python knowing in advance how much space to allocate for the new string.

openformats/formats/indesign.py
```diff
@@ -0,0 +1,181 @@
...
+            try:
+                story_content = idml[key]
+            except KeyError:
+                continue
+
+            story_content = idml[key]
+            transcriber = Transcriber(story_content)
+            found = True
+            while found:
+                try:
```
kbairak: Minor: When we have something like this:

```python
try:
    a()
    b()  # May raise some kind of exception
    c()
    d()
except SomeKindOfException:
    e()
```

I prefer it to be like:

```python
try:
    a()
    b()  # May raise some kind of exception
except SomeKindOfException:
    e()
else:
    c()
    d()
```

I think it makes reading the code a bit easier.

SofiaMargariti:
> 2. What happens when a translation has been removed from the stringset? (...) In this case, if a string was missing from the stringset, I think the whole <Content> tag should be removed. We can let this pass if we know that the rest of the InDesign software stack will crash if <Content> tags are missing from the "stories".

@kbairak you are right we shouldn't depend on txc for this, but I think it is more valid to keep the <Content> tags and just replace its text with an empty string. This way the file can be re-uploaded again to Transifex as translation file for the same resource.

openformats/formats/indesign.py
```diff
@@ -0,0 +1,204 @@
...
+            current_string: the last unprocessed string to be passed to the
+                            next story
+        """
+        transcriber = Transcriber(story_content)
+        hash_regex = re.compile('[a-z,0-9]{32}_tr')
+        found = True
+        while found:
+            try:
+                if not current_string:
+                    current_string = next(stringset)
```
kbairak: Lets suppose this story here doesn't have any hashes that match the stringset. We extract the next string from the stringset, don't replace it with anything (`ValueError` on the `index` call). Then, when compiling the next story, that string will be missing. Although I think we've faced this before, I don't recall what we did specifically. Perhaps we probably need some kind of "put back" methodology, eg:

```python
class Stringset(list):
    def next(self):
        return self.pop(0)

    def put_back(self, string):
        return self.insert(0, string)
```

or simply store a copy of `stringset` in `self.stringset` and call the `pop` and `insert` methods directly.

SofiaMargariti: @kbairak the `current_string` here is supposed to play this role. Because I am working with an iterator and there is no way (that I know of at least) to rewind it back. I am keeping the string we were processing when the ValueError was raised, return it and pass it to the next call of the _compile_story. 
But working with a list might probably be clearer now that we've split it up.

PR #87, opened 2018-02-15
InDesign finalizations

rigaspapas:
Related to [TX-9144: InDesign's finalizations](https://transifex.atlassian.net/browse/TX-9144)

Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------
1. Empty strings appeared in the editor (when only non-printable characters were included)
2. Spaces around the text where not preserved

Steps to reproduce
------------------
1. Use the ifood IDML file from this [doc](https://docs.google.com/document/d/1XEkdNjt5JCp-_shSTQ775L8rvnstrMA5-QiGsW1qHg4/edit)
2. Use the TrustYouBranding file from the same doc and the `recent research results` string

Solution
--------
1. Ignore strings that don't contain letters, symbols or punctuation (as defined [here](https://www.unicode.org/reports/tr44/#General_Category_Values))
2. Extend the regular expression to ignore wrapping spaces

### DIFF
--- a/openformats/formats/indesign.py
+++ b/openformats/formats/indesign.py
@@ -26,7 +26,9 @@ class InDesignHandler(Handler):
     SPECIFIER = None
     PROCESSES_BINARY = True
 
-    CONTENT_REGEX = re.compile(r'(<Content>)(.*)?(</Content>)')
+    # The ? at the end of the string regex, makes it non-greedy in order to
+    # allow trailing spaces to be preserved
+    CONTENT_REGEX = re.compile(r'(<Content>\s*)(.*?)(\s*</Content>)')
     SPECIAL_CHARACTERS_REGEX = re.compile(r'<\?ACE \d+\?>|<Br/>;')
 
     """ Parse Methods """
--- a/openformats/formats/indesign.py
+++ b/openformats/formats/indesign.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import io
 import re
+import unicodedata
 
 from itertools import count
 from lxml import etree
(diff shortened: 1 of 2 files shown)


openformats/formats/indesign.py
```diff
@@ -104,6 +104,9 @@ def _can_skip_content(self, string):
             return True
         except ValueError:
             pass
+        # Special content in BackingStory.xml
+        if u'\ufeff' == string:
+            return True
```
kouk: @rigaspapas why not just check for content after stripping all non-printable characters?
ref: https://mayart.de/download/Indesign-IDML/special-idml-chars.pdf

one idea would be to use the `unicodedata` module to ignore all character not in printable categories:
https://www.unicode.org/reports/tr44/#General_Category_Values

kouk: The reason I'm saying this is who's to say we don't find another file that e.g. has a `u'\u200c'` character?

rigaspapas: @kouk what we want to achieve is to avoid exporting text that is not translatable. There is no standard way in Unicode characters to distinguish letters from symbols. Is there any?

You are right that any non-printable character can cause the same unwanted result. I considered ignoring one specific character, because I guess that's what InDesign insert in every document automatically (not a user input).

openformats/tests/formats/indesign/test_indesign.py
```diff
@@ -44,16 +44,22 @@ def test_can_skip_content(self):
...
-            '   <?ACE 7?> ',
-            '',
-            '<?ACE 8?> <Br/>;',
+            u' ',
+            u'   <?ACE 7?> ',
+            u'',
+            u'<?ACE 8?> <Br/>;',
+            u'\ufeff',
+            u' \ufeff ',
+            u' \ufeff 5',
```
kouk: here's another test case: 
```
u'<Br/>\ufeff'
```
;-)

openformats/formats/indesign.py
```diff
@@ -104,6 +107,22 @@ def _can_skip_content(self, string):
...
+        translated. We assume that translatable characters are the letters,
+        the symbols and the punctuation.
+        """
+        acceptable = ["L", "P", "S"]
+
+        for letter in string:
+            char_type = unicodedata.category(letter)
+            if char_type[0] in acceptable:
+                return True
         return False
```
kouk: just for fun:
```
from six.moves import map
return any(c[0] in ["L", "P", "S"] for c in map(unicodedata.category, string))
```
or even
```
from six.moves import map
return any(map(["L", "P", "S"].__contains__,
               map(itemgetter(0), map(unicodedata.category, string))) 
```

openformats/tests/formats/indesign/test_indesign.py
rigaspapas: You are right. We should first strip the special characters and then check for the translatable ones. Thanks!

openformats/formats/indesign.py
rigaspapas: @kouk I really like the first approach, which is very *functional*, but it would require a new dependency. So, should we leave it as is?

kouk: sure, it's fine like it is. But these examples don't really require the new dependency, you could do `from itertools import imap`. It's just that this way it's compatible with both python2 and python3.

PR #88, opened 2018-02-23
[DNM] Yaml standalone

SofiaMargariti:
* Make YamlHandler currently used by GithubMarkdownV2Handler a standalone parser.
* Make it aware of pluralized strings and add hooks in order subclasses to be able to handle pluralized strings. (A first version of how the hooks might look like is in https://github.com/transifex/openformats/tree/i18n_yaml where the i18n specific handler will be implemented).

Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------

Steps to reproduce
------------------

Solution
--------

Example run / Screenshots
-------------------------

Performance on live data
------------------------

### DIFF
--- a/openformats/formats/yaml.py
+++ b/openformats/formats/yaml.py
@@ -14,7 +14,7 @@ from ..transcribers import Transcriber
 
 class YamlHandler(Handler):
     name = "Yaml"
-    extension = ".yml"
+    extension = "yml"
     EXTRACTS_RAW = False
 
     BACKSLASH = u'\\'
--- a/openformats/formats/yaml.py
+++ b/openformats/formats/yaml.py
@@ -19,12 +19,14 @@ class YamlHandler(Handler):
 
     BACKSLASH = u'\\'
     DOUBLE_QUOTES = u'"'
+    SINGLE_QUOTES = u'\''
     NEWLINE = u'\n'
     COLON = u':'
     ASTERISK = u'*'
     AMPERSAND = u'&'
     DASH = u'-'
     HASHTAG = u'#'
+    BACKTICK = u'`'
 
     def _should_wrap_in_quotes(self, tr_string):
         return any([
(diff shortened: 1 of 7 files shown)


openformats/formats/yaml.py
```diff
@@ -204,12 +224,47 @@ def _find_comment(self, content, start, end):
         else:
             return None
 
+    def get_yaml_data_to_parse(self, yaml_data):
+        """ Returns the part of the yaml_data that actually need to be parsed
+
+        In some cases part only a part of the YAML structure needs to be
```
rigaspapas: There is a typo here: "part only a part"

openformats/tests/formats/yaml/files/1_en_exported.yml
```diff
@@ -0,0 +1,36 @@
+---
```
rigaspapas: I am not sure this file is used. We need to check this out before we add it.

rigaspapas:
I think there is a problem with the expression we use to decide whether we need to wrap something in quotes.

Take this as an example:
```
electron_is_easy:
  title: test
```

If you parse it, and provide a translation that starts with a quote or backtick, then the produced (compiled) YML file is not parseable:
```
electron_is_easy:
  title: 'σε μας
```

Traceback:
```Traceback (most recent call last):
  File "...../openformats/testbed/main/views.py", line 102, in _parse
    template, stringset = handler.parse(source)
  File "...../openformats/openformats/formats/yaml.py", line 264, in parse
    yaml_data = self._load_yaml(content, loader=TxYamlLoader)
  File "...../openformats/openformats/formats/yaml.py", line 76, in _load_yaml
    raise ParseError(unicode(e))
ParseError: while scanning a quoted scalar
  in "<unicode string>", line 2, column 10
found unexpected end of stream
  in "<unicode string>", line 2, column 17```

PR #93, opened 2018-03-21
I18n yaml

SofiaMargariti:
Checklist (for the reviewer)
----------------------------

* [x] Code is well documented
* [x] Proper labels have been applied

Problem
-------
Create a new version of YAML parser for internationalization specific YAML files.

Steps to reproduce
------------------
Either though TX app pointing to this branch try to upload a yaml file or using the testbed.

Solution
--------
Create YamlHandler (introduced in #88) subclass to handle internationalization specific YAML files (extra handling needed for root key and pluralized strings).

In the process there where some bugs fixed in the in the original YamlHandler class and the compilation logic was completely rewritten.

Thinks left to be done:
* [x] set `should_use_template` based on mode
* ~~respect style for entries of pluralized strings~~ (this doesn't actually make sense because different languages can have different number of plural rules, we don't have styles for all of them)
* [x] more tests and documentation
* [x] squash commits

Example run / Screenshots
-------------------------

Performance on live data
------------------------

### DIFF
--- a/openformats/formats/yaml.py
+++ b/openformats/formats/yaml.py
@@ -14,7 +14,7 @@ from ..transcribers import Transcriber
 
 class YamlHandler(Handler):
     name = "Yaml"
-    extension = ".yml"
+    extension = "yml"
     EXTRACTS_RAW = False
 
     BACKSLASH = u'\\'
--- a/openformats/formats/yaml.py
+++ b/openformats/formats/yaml.py
@@ -19,12 +19,14 @@ class YamlHandler(Handler):
 
     BACKSLASH = u'\\'
     DOUBLE_QUOTES = u'"'
+    SINGLE_QUOTES = u'\''
     NEWLINE = u'\n'
     COLON = u':'
     ASTERISK = u'*'
     AMPERSAND = u'&'
     DASH = u'-'
     HASHTAG = u'#'
+    BACKTICK = u'`'
 
     def _should_wrap_in_quotes(self, tr_string):
         return any([
@@ -34,6 +36,9 @@ class YamlHandler(Handler):
             tr_string.lstrip().startswith(self.ASTERISK),
             tr_string.lstrip().startswith(self.AMPERSAND),
             tr_string.lstrip().startswith(self.DASH),
+            tr_string.lstrip().startswith(self.SINGLE_QUOTES),
+            tr_string.lstrip().startswith(self.BACKTICK),
+            tr_string.rstrip().endswith(self.BACKTICK),
         ])
 
     def _wrap_in_quotes(self, tr_string):
--- a/openformats/formats/yaml.py
+++ b/openformats/formats/yaml.py
@@ -9,9 +9,54 @@ from yaml.constructor import ConstructorError
 from ..handlers import Handler
 from ..strings import OpenString
 from ..exceptions import ParseError
+from ..transcribers import Transcriber
 
 
 class YamlHandler(Handler):
+    name = "Yaml"
+    extension = ".yml"
+    EXTRACTS_RAW = False
+
+    BACKSLASH = u'\\'
+    DOUBLE_QUOTES = u'"'
+    NEWLINE = u'\n'
+    COLON = u':'
+    ASTERISK = u'*'
+    AMPERSAND = u'&'
+    DASH = u'-'
+    HASHTAG = u'#'
+
+    def _should_wrap_in_quotes(self, tr_string):
+        return any([
+            self.NEWLINE in tr_string[:-1],
+            self.COLON in tr_string,
+            self.HASHTAG in tr_string,
+            tr_string.lstrip().startswith(self.ASTERISK),
+            tr_string.lstrip().startswith(self.AMPERSAND),
+            tr_string.lstrip().startswith(self.DASH),
+        ])
+
+    def _wrap_in_quotes(self, tr_string):
+        if self._should_wrap_in_quotes(tr_string):
+            # escape double quotes inside strings
+            tr_string = tr_string.replace(
+                self.DOUBLE_QUOTES,
+                (self.BACKSLASH + self.DOUBLE_QUOTES)
+            )
+            # surround string with double quotes
+            tr_string = (self.DOUBLE_QUOTES + tr_string +
+                         self.DOUBLE_QUOTES)
+        return tr_string
+
+    def _compile_single(self, string):
+        tr_string = self._wrap_in_quotes(string.string)
+        # this is to ensure that if the style is literal or folded
+        # http://www.yaml.org/spec/1.2/spec.html#id2795688
+        # a new line always follows the string
+        if (string.flags and string.flags in '|>' and
+                tr_string[-1] != self.NEWLINE):
+            tr_string = tr_string + self.NEWLINE
+        return tr_string
 
     def _load_yaml(self, content, loader):
         """
@@ -30,7 +75,8 @@ class YamlHandler(Handler):
         except yaml.scanner.ScannerError as e:
             raise ParseError(unicode(e))
 
-    def _parse_leaf_node(self, yaml_data, parent_key, style=[]):
+    def _parse_leaf_node(self, yaml_data, parent_key, style=[],
+                         pluralized=False):
         """Parse a leaf node in yaml_dict.
         Args:
             yaml_data: A tuple of the form (string, start, end, style)
@@ -42,7 +88,8 @@ class YamlHandler(Handler):
         Returns:
             A dictionary representing the parsed leaf node
         """
-        value = yaml_data[0]
+        val = yaml_data[0]
+        value = self.parse_pluralized_value(val) if pluralized else val
         start = yaml_data[1]
         end = yaml_data[2]
         style.append(yaml_data[3] or '')
@@ -51,11 +98,14 @@ class YamlHandler(Handler):
             'end': end,
             'key': parent_key,
             'value': value,
-            'style': ':'.join(style or [])
+            'style': ':'.join(style or []),
+            'pluralized': pluralized,
         }
 
     @staticmethod
     def unescape_dots(k):
+        """ We use dots to construct the strings key, so we need to
+        escape dots that are part of an actual YAML key """
         return k.replace('.', '<TX_DOT>')
 
     def _get_key_for_node(self, key, parent_key):
@@ -103,8 +153,20 @@ class YamlHandler(Handler):
             for key, val in yaml_data.items():
                 node_key = self._get_key_for_node(key, parent_key)
                 style = copy.copy(parent_style)
-                if isinstance(val[0], (dict, list)):
-                    # style.append(val.get('style', ''))
+                if isinstance(val[0], dict):
+                    if self.is_pluralized(val[0]):
+                        parsed_data.append(
+                            self._parse_leaf_node(
+                                val, node_key,
+                                style=copy.copy(parent_style or []),
+                                pluralized=True
+                            )
+                        )
+                    else:
+                        parsed_data = self.parse_yaml_data(
+                            val[0], node_key, parsed_data, context,
+                            parent_style=copy.copy(style or []))
+                elif isinstance(val[0], list):
                     parsed_data = self.parse_yaml_data(
                         val[0], node_key, parsed_data, context,
                         parent_style=copy.copy(style or []))
@@ -119,15 +181,15 @@ class YamlHandler(Handler):
             # using the position (index) of it as parent key using
             # brackets around it. I.e.: 'foo.[0].bar'.
             for i, e in enumerate(yaml_data):
-                p_key = parent_key + '.[%s]' % (i)
+                node_key = self._get_key_for_node('[%s]' % (i), parent_key)
                 if isinstance(e, (dict, list)):
                     parsed_data = self.parse_yaml_data(
-                        e, p_key, parsed_data, context,
+                        e, node_key, parsed_data, context,
                         parent_style=copy.copy(parent_style or []))
                 else:
                     parsed_data.append(
                         self._parse_leaf_node(
-                            e, p_key, style=copy.copy(parent_style or [])
+                            e, node_key, style=copy.copy(parent_style or [])
                         )
                     )
         else:
@@ -162,12 +224,47 @@ class YamlHandler(Handler):
         else:
             return None
 
+    def get_yaml_data_to_parse(self, yaml_data):
+        """ Returns the part of the yaml_data that actually need to be parsed
+
+        In some cases only a part of the YAML structure needs to be
+        processed for strings to be extracted. e.g. when the root key is the
+        language code we need to exclude this from the rest of the processing.
+        In the generic case the whole structure will be processed.
+        """
+        return yaml_data
+
     def parse(self, content, **kwargs):
+        """ Parses the given YAML content to create stringset and template
+
+        Steps are:
+            1. Load yaml content using our custom loader TxYamlLoader that
+               in addition to the value for each key notes the `start` and
+               `end` index of each node in the file and some metadata.
+            2. Flattens the output of the loader to be a list of the form:
+               ```
+               [{
+                   'key': 'string_key1',
+                   'value': 'string1',
+                   'end': <end_index_of_node>,
+                   'start': <start_index_value>,
+                   'style': '|, >, ...'
+                },
+                ...
+               ]
+               ```
+            3. Iterates over the flattened list and for each entry creates an
+               OpenString object, appends it to stringset and replace its value
+               with the template_replacement in the template.
+            4. Returns the (template, stringset) tuple.
+        """
         template = ""
         context = ""
         stringset = []
         yaml_data = self._load_yaml(content, loader=TxYamlLoader)
-        parsed_data = self.parse_yaml_data(yaml_data, '', [], context)
+        yaml_data = self.get_yaml_data_to_parse(yaml_data)
+        parsed_data = self.parse_yaml_data(yaml_data, '', [],
+                                           context)
         parsed_data.sort()
 
         end = 0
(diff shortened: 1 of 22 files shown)


openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+            raise ParseError(
+                "Expected a mapping node, but found %s" % node.id,
+            )
+        pairs = []
+        for key_node, value_node in node.value:
+            try:
+                key = self.construct_object(key_node, deep=deep)
+                value = self.construct_object(value_node, deep=deep)
+            except ConstructorError, e:
+                print("During parsing YAML file: %s" % unicode(e))
```
dbendilas: --> `except ConstructorError as e` to be compatible with Python 3.6. Please check for other similar occurrences in all new/edited `.py` files.

openformats/formats/yaml/yaml.py
```diff
@@ -280,60 +250,160 @@ def parse(self, content, **kwargs):
...
-            if isinstance(value, dict) or value:
-                string_object = OpenString(
-                    key, value, context=context, flags=style, order=order,
-                    comment=comment
-                )
-                stringset.append(string_object)
-                order += 1
-                template += (content[end:start] +
-                             string_object.template_replacement)
+            comment = ''
```
dbendilas: The initial value of `comment` is never used, can be omitted altogether.

openformats/formats/yaml/yaml.py
```diff
@@ -168,10 +139,12 @@ def parse_yaml_data(self, yaml_data, parent_key, parsed_data,
                             )
                         )
                     else:
+                        style.append(val[3] or '')
```
dbendilas: (I cannot add my comment on the exact line I want, i.e. line 113, because it wasn't edited on this PR)

Mutable objects shouldn't be used as default values (`parent_style` attribute on `def parse_yaml_data()`).
In short, default values are initialized only once when defined and the same instance is used from then on.
This can result in unexpected behavior and bugs that are really hard to debug.

You can replace this with:

```
def parse_yaml_data(self, yaml_data, parent_key, parsed_data,
                         context="", parent_style=None):
    parent_style = parent_style or []
```

See http://docs.python-guide.org/en/latest/writing/gotchas/ for more.

openformats/formats/yaml/yaml.py
```diff
@@ -1,67 +1,41 @@
...
+    # number of spaces used for indentation in YAML content
+    indent = 2
+
+    # the levels of indentation of a string can be extracted by doing
+    # `len(keys.split('.'))`. For files that we ignore the root key
+    # (e.g ruby i18n files) we need to set an extra level of indentation
+    # `extra_indent = 1`
+    extra_indent = 0
+
+    # When compiling for a mode that needs needs to completely remove
```
dbendilas: `needs` repeats twice.

openformats/formats/yaml/yaml.py
```diff
@@ -280,60 +250,160 @@ def parse(self, content, **kwargs):
...
+            emitter.write_literal(string.string)
+            translation = emitter.stream.getvalue()
+        elif style == '>':
+            emitter.write_folded(string.string)
+            translation = emitter.stream.getvalue()
+
+        emitter.stream.close()
+        return translation
+
+    def _apply_translations(self, template, stringset, **kwargs):
```
dbendilas: I think a more descriptive name would be better, e.g. `_compile_from_template()`.

openformats/formats/yaml/yaml.py
```diff
@@ -280,60 +250,160 @@ def parse(self, content, **kwargs):
...
+        if self.should_use_template:
+            return self._apply_translations(template, stringset)
+        else:
+            yg = YamlGenerator(self)
+            yaml_dict = yg._generate_yaml_dict(stringset)
+            if self.language:
+                yaml_dict = self._wrap_yaml_dict(yaml_dict, self.language)
+            return yaml.dump(yaml_dict, width=float('inf'),
+                             Dumper=TxYamlDumper, allow_unicode=True,
+                             indent=self.indent).decode("utf-8")
```
dbendilas: For better clarity, I suggest that you also create another method, e.g. `_compile_without_template()`, and move the code from `else:` to the new method.

openformats/formats/yaml/yaml.py
dbendilas: Also, what exactly is `val`, `val[0]`, `val[3]` and so on? Some context in the docstring or a comment would be very helpful for future debugging or feature implementations.

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+
+class TxYamlLoader(yaml.SafeLoader):
+    """
+    Custom YAML Loader for Tx
+    """
+    def __init__(self, *args, **kwargs):
+        super(TxYamlLoader, self).__init__(*args, **kwargs)
+        self.stream = args[0]
+        self.post_block_comment_pattern = re.compile(
+            r'(?:#.*\r?\n\s*)+$')
```
rigaspapas: This fits in one line...

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+        content = self.stream[start:end]
+        m = self.post_block_comment_pattern.search(content)
+        if m:
+            end = start + m.start()
+        return end
+
+    def construct_mapping(self, node, deep=True):
+        if not isinstance(node, yaml.MappingNode):
+            raise ParseError(
+                "Expected a mapping node, but found %s" % node.id,
```
rigaspapas: This one fits in one line as well. Was the trailing comma intended here?

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+        Returns:
+            An integer for the new end position.
+        """
+        content = self.stream[start:end]
+        m = self.post_block_comment_pattern.search(content)
+        if m:
+            end = start + m.start()
+        return end
+
+    def construct_mapping(self, node, deep=True):
```
rigaspapas: Is this method ever called? If we override it in order to alter the library's default behaviour, we need to explain why we do this, either in the docstring or in comments.

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+        return obj
+
+    def _insert_translation_in_dict(self, parent_dict, keys, styles,
+                                    translation_string):
+        """ Given a set of keys and a set of styles generate a Yaml
+        serializable dictionary with each nested element casted to the
+        right yaml representee class.
+
+        Args:
+            keys: a list procuded by the source string key splitted on `.`
```
rigaspapas: Typos: procuded -> produced & splitted -> split (split is irregular)

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+            OrderedDict([
+                (u'one', OrderedDict([
+                    (u'two', BlockList([
+                        single_quoted_unicode(u'test')
+                    ]))
+                ]))
+            ])
+
+            # re-calling _insert_translation_in_dict with second set of keys
+            # the `result` as `parent_dict` whould update the same dictionary
```
rigaspapas: typo: whould -> would

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+        elif style == '"':
+            obj = double_quoted_unicode(obj)
+        elif style == "'":
+            obj = single_quoted_unicode(obj)
+        return obj
+
+    def _insert_translation_in_dict(self, parent_dict, keys, styles,
+                                    translation_string):
+        """ Given a set of keys and a set of styles generate a Yaml
+        serializable dictionary with each nested element casted to the
```
rigaspapas: casted -> cast

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
+from __future__ import absolute_import
```
rigaspapas: This file contains both tabs and spaces. Indentation in docstrings is not consistent

rigaspapas:
A general comment for this PR. I see that we use both `YAML`, `Yaml` and `yaml`. We need to stick with one style in order to be consistent.

openformats/formats/yaml/yaml.py
```diff
@@ -280,60 +250,160 @@ def parse(self, content, **kwargs):
...
-                stringset.append(string_object)
-                order += 1
-                template += (content[end:start] +
-                             string_object.template_replacement)
+            comment = ''
+            if not value:
+                continue
+            if isinstance(value, dict) and not all(value.values()):
+                continue
+            string_object = OpenString(
```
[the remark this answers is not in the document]
SofiaMargariti: Here value will be either a string or a dictionary similar to: `{1: "singular string", 5: "pluralized string"}`, so we don't need to check for a list.

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+                        if styles[i] == 'block':
+                            yaml_dict.append(BlockStyleOrderedDict())
+                    yaml_dict = yaml_dict[index]
+            if is_last:
+                if next_is_list:
+                    yaml_dict.append(self._get_styled_string(entry,
+                                                             styles[-1]))
+                else:
+                    yaml_dict[next_key] = self._get_styled_string(entry,
+                                                                  styles[-1])
```
dbendilas: This method seems too complex, not very easy to read and debug.

I think something like the following is better, as it separates the high-level logic with the logic for the creation of the proper node instance, and includes comments in a few key places:

WARNING: Untested code

```
def _insert_translation_in_dict(self, parent_dict, keys, styles,
                                translation_string):
    yaml_dict = parent_dict
    entry = translation_string
    if len(keys) == 1:
        key = self._parse_int_key(keys[0])
        yaml_dict[key] = self._get_styled_string(entry, styles[-1])
        return

    for i, key in enumerate(keys[:-1]):
        is_last = i == (len(keys) - 2)
        next_key = keys[i+1]

        key = self._parse_int_key(key)
        key, is_list = self._parse_list_index_key(key)
        next_key, next_is_list = self._parse_list_index_key(next_key)

        if is_list:
            index = key
            if index < len(yaml_dict):
                yaml_dict = yaml_dict[index]
            else:
                if next_is_list:
                    node = self._create_list_node(styles[i])
                else:
                    node = self._create_dict_node(styles[i])
                yaml_dict.append(node)
                yaml_dict = yaml_dict[index]

        else:
            if key in yaml_dict:
                yaml_dict = yaml_dict.get(key)
            else:
                # Comment goes here
                if next_is_list:
                    node = self._create_list_node(styles[i + 1])
                # Comment goes here
                else:
                    if is_last:
                        node = self._create_dict_node(styles[i])
                    else:
                        node = self._create_dict_node()
                yaml_dict[key] = node
                yaml_dict = yaml_dict[key]

        # Comment goes here
        if is_last:
            if next_is_list:
                yaml_dict.append(self._get_styled_string(entry,
                                                            styles[-1]))
            else:
                yaml_dict[next_key] = self._get_styled_string(entry,
                                                                styles[-1])

def _create_list_node(self, style):
    if style == 'block':
        return BlockList()
    elif style == 'flow':
        return FlowList()

def _create_dict_node(self, style=None):
    if style == 'block':
        return BlockStyleOrderedDict()
    elif style == 'flow':
        return FlowStyleOrderedDict()

    return OrderedDict()
```

openformats/formats/yaml/yaml_representee_classes.py
```diff
@@ -0,0 +1,40 @@
...
+    pass
+
+
+class double_quoted_unicode(unicode):
+    pass
+
+
+class single_quoted_unicode(unicode):
+    pass
+
```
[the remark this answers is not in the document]
SofiaMargariti: They were like that in TXC and I just carried them over. My guess is because they subclass `unicode` which is lowercase.

openformats/formats/yaml/utils.py
SofiaMargariti: this was also carried over.. I have changed it to use `format` anyway

openformats/formats/yaml/yaml_representee_classes.py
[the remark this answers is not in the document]
dbendilas: Duh, you are right, I didn't notice.

openformats/formats/yaml/yaml.py
SofiaMargariti: @dbendilas I did a refactoring to use an `namedtuple`, so now instead of the index you can access each element of `val` with a name and renamed `val` to `node`. 
```
val[0] -> node.value
val[1] -> node.start
val[2] -> node.end
val[3] -> node.style
```

openformats/formats/yaml/utils.py
SofiaMargariti: @rigaspapas I couldn't find any tabs in this file. Pls check that this is still an issue.

SofiaMargariti:
@dbendilas @rigaspapas I believe I have addressed all of your comments, pls take another look when you find the time.

dbendilas:
@SofiaMargariti I liked the use of nodetuple!

openformats/tests/formats/yaml/files/1_en.yml
```diff
@@ -35,6 +37,19 @@ key.with.dots: dot value
...
+
+<weird_key>: value
+"[weird_key]": value
+
+true: boolean key
+
+double_list:
+  - - test
+
+double_flow_list: [[ test ]]
```
dbendilas: I don't see any entry with a simple flow list with multiple items.  I think we should have one.

openformats/formats/yaml/yaml_representee_classes.py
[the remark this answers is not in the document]
rigaspapas: @SofiaMargariti I have no problem to keep the class names, but we should not leave flake8 warnings in the open:
`class folded_unicode(unicode):  # noqa: N801`

openformats/tests/formats/yaml/files_i18n/1_en.yml
```diff
@@ -0,0 +1,10 @@
+en:
+  normal string: normal string
+  pluralized_string:
+    one: One
+    other: Other
+  another normal string: normal string
+
+  empty_pluralized_string:
+    one: ""
+    other: ""
```
dbendilas: We should add all the features of the base handler here too, otherwise we cannot be sure that it supports everything it needs to.

openformats/tests/formats/yaml/test_yaml.py
```diff
@@ -11,11 +13,123 @@ class YamlTestCase(CommonFormatTestMixin, unittest.TestCase):
...
         """Test that import-export is the same as the original file."""
         remade_orig_content = self.handler.compile(self.tmpl, self.strset)
         self.assertEquals(remade_orig_content, self.data["1_en_exported"])
+
+    def test_compile_with_template(self):
+        """Test that import-export is the same as the original file."""
+        self.handler.should_use_template = False
+        remade_orig_content = self.handler.compile(self.tmpl, self.strset)
+        self.assertEquals(remade_orig_content,
+                          self.data["1_en_exported_without_template"])
```
dbendilas: Shouldn't this method be named `test_compile_without_template`?

openformats/formats/yaml.py
```diff
@@ -1,463 +0,0 @@
...
-    NEWLINE = u'\n'
-    COLON = u':'
-    ASTERISK = u'*'
-    AMPERSAND = u'&'
-    DASH = u'-'
-    HASHTAG = u'#'
-    BACKTICK = u'`'
-
-    def _should_wrap_in_quotes(self, tr_string):
-        return any([
```
dbendilas: @SofiaMargariti sorry, I just remembered one more thing. I think it would be really good if the main public methods go to the top, i.e. `parse()` and `compile()`. Right now it's not easy to navigate the class.

openformats/formats/yaml/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .yaml import YamlHandler
+from .yaml_i18n import I18nYamlHandler
```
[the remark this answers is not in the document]
SofiaMargariti: So that someone can import handlers like this: `opeformats.formats.yaml.I18nYamlHandler` instead of `opeformats.formats.yaml_i18n.I18nYamlHandler`. I don't think that anybody needs to care about the internal structure of the `yaml/` folder, just that it exposes the two handlers to the public. All other utils and classes are private to folder and used only by the two main classes.

dbendilas:
The new changes look good! @SofiaMargariti

openformats/formats/yaml/utils.py
```diff
@@ -0,0 +1,365 @@
...
+            ```
+        """
+        yaml_dict = parent_dict
+        entry = translation_string
+        if len(keys) == 1:
+            key = self._parse_int_key(keys[0])
+            yaml_dict[key] = self._get_styled_string(entry, styles[-1])
+            return
+
+        for i, key in enumerate(keys[:-1]):
```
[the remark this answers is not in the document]
SofiaMargariti: @dbendilas we are processing the last key together with the value here: https://github.com/transifex/openformats/blob/adeeac28c5fcd48064a67ed42c3a56969a0ced23/openformats/formats/yaml/utils.py#L342.

This way we can skip it from the loop and also we are sure that `next_key` will always exists, so we avoid one extra check.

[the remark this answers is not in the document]
dbendilas: Thanks @SofiaMargariti 👍

openformats/formats/yaml/yaml_i18n.py
```diff
@@ -53,9 +53,9 @@ def _compile_pluralized(self, string):
         indent = " " * indentation_levels * self.indent
 
         yml_str = yaml.dump(plurals_dict, default_flow_style=False,
-                            allow_unicode=True)
-        plural_entry = ''.join([
-            "{indent}{line}".format(indent=indent, line=line)
+                            allow_unicode=True).decode('utf-8')
+        plural_entry = u''.join([
+            u"{indent}{line}".format(indent=indent, line=line)
```
rigaspapas: Should we add a testcase that breaks the old code?

openformats/tests/formats/yaml/files/1_en_exported_without_template.yml
```diff
@@ -32,3 +32,5 @@ double_list:
   - - test
 simple_flow_list: ["one", 'two', three]
 double_flow_list: [[test]]
+anchor_key:
+  - value
```
rigaspapas: Why id the `alias_key` omitted here?

openformats/tests/formats/yaml/files/1_en.yml
```diff
@@ -65,3 +65,8 @@ binary_value: !!binary |
 ? - Manchester United
   - Real Madrid
 : [ 2001-01-01, 2002-02-02 ]
+
+# anchors and aliases
+anchor_key: &an_anchor
+  - value
+alias_key: *an_anchor
```
rigaspapas: Nice. Should we another value to `alias_key` to test the mix of the aliases with other values?
For example:
```
alias_key:
    - *an_anchor
    - another true value
```

openformats/formats/yaml/yaml_i18n.py
SofiaMargariti: I updated the test files for i18n handler to include some non unicode characters.

PR #98, opened 2018-04-18
Relax condition for pluralized yaml nodes

SofiaMargariti:
Checklist (for the reviewer)
----------------------------

* [x] Code is well documented

Problem
-------
Right we expect a pluralized node to have exactly the plural rules the language supports and no extra key, otherwise we don't consider it pluralized. We need to relax this condition to allow extra keys to exist as long as all the plural rules are there.

Steps to reproduce
------------------
See updated test files for examples.

Solution
--------
* Consider a node pluralized even if there are extra keys in addition to the language plural rule keys as long as they are valid plural rules. Extra plural rules in this case are ignored completely, they don't appear in the editor and are not preserved in the translation file. If the extra keys are not plural rules the node is parsed as non-pluralized and each key will be a separate string.
* Make sure that when one or more of the plural rule keys contain a complex value, i.e. not a string, then this node is not going to be pluralized.

Example run / Screenshots
-------------------------

Performance on live data
------------------------

### DIFF
--- a/openformats/tests/formats/yamlinternationalization/files/1_en.yml
+++ b/openformats/tests/formats/yamlinternationalization/files/1_en.yml
@@ -9,6 +9,19 @@ en:
     one: ""
     other: ""
 
+  pluralized_string_with_extra_keys:
+    zero: Zero
+    one: One
+    other: Other
+  non_pluralized_string_with_extra_keys:
+    extra_key: random value
+    one: One
+    other: Other
+  plural_key_with_nested_nodes:
+    one:
+      nested_key: One
+    other: Other
+
   title: About writing and formatting on GitHub
   intro: into text
   numeric_var: 12.5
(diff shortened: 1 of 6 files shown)


openformats/tests/formats/yamlinternationalization/files/1_en.yml
```diff
@@ -9,6 +9,18 @@ en:
     one: ""
     other: ""
 
+  pluralized_string_with_extra_keys:
+    one: One
+    other: Other
```
dbendilas: Shouldn't `pluralized_string_with_extra_keys` contain at least 1 extra key?

SofiaMargariti: yes, I had added it.. I probably did something wrong in the rebases. I am fixing it now.

PR #99, opened 2018-04-18
Fix testbed sample support for YML i18n

dbendilas:
Problem
-------
Before, the existing sample file for yml i18n did not appear on the testbed when selecting this format; instead the textarea remained empty. Now, it appears normally, just as it happens for other formats.

Steps to reproduce
------------------
Run the testbed and select 'Yaml (Internationalization)' from the dropdown. The content in the textarea should be populated with the yaml i18n sample, but it's not.

Solution
--------
In order to fix this, some files had to be moved. This was necessary because of the existing naming conventions that are in place for the folder structure of the test template files.


openformats/tests/formats/yaml (internationalization)/test_i18n_yaml.py
```diff
@@ -7,7 +7,8 @@
 
 class I18nYamlTestCase(CommonFormatTestMixin, unittest.TestCase):
     HANDLER_CLASS = I18nYamlHandler
-    TESTFILE_BASE = "openformats/tests/formats/yaml/files_i18n"
+    TESTFILE_BASE = "openformats/tests/formats/" \
+                    "yaml (internationalization)/files"
```
rigaspapas: Why did you chose `yaml (internationalization)` over `yaml_i18n`?

dbendilas: @rigaspapas I didn't choose it. I had to use this, because this is the name defined in https://github.com/transifex/openformats/blob/252a9af8978977bcf4d1fa4b3f40f2b134dc7640/openformats/formats/yaml/yaml_i18n.py#L16

and the dynamic folder structure needs to use the lowercase version of that name, as seen in https://github.com/transifex/openformats/blob/252a9af8978977bcf4d1fa4b3f40f2b134dc7640/testbed/main/views.py#L85

PR #100, opened 2018-04-20
Test directories now contain only letters and numbers

SofiaMargariti:
Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------
After the changes in #99 it seems that some test cases cannot be discovered by node and as a result they don't run.

Steps to reproduce
------------------
Run test suit and see that I18nYamlTestCase does not run (maybe by adding a debugger in the setUp method).

Solution
--------
Make handler specific folder names only lowercase letters.
If we want files to be discovered by testbed the folder names should be lowercase version of the Handler name with all non letter characters removed.

Example run / Screenshots
-------------------------

Performance on live data
------------------------


testbed/main/views.py
```diff
@@ -81,7 +81,7 @@ def post(self, request):
     def _choose_handler(self, payload):
         handler_name = payload['handler']
         handler_class = self.handlers[handler_name]
-        handler_name_lower = handler_name.lower()
+        handler_name_lower = filter(unicode.isalpha, handler_name).lower()
         sample_filepath = os.path.join(
```
dbendilas: @SofiaMargariti why not `isalnum`? That would be more future-proof.

SofiaMargariti: You are right, I fixed it!

PR #101, opened 2018-04-23
Support custom tags in YAML format

rigaspapas:
Extend the YAML parser in order to include any translatable content that is marked with a custom type using tags.

Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------

YAML parser ignores any content behind [custom tags](http://camel.readthedocs.io/en/latest/yamlref.html#tags)

Steps to reproduce
------------------

Upload this YAML file:
```
foo: !bar Hello World
```

The translatable content `Hello World` will be ignored.

Solution
--------

The solution consists of two parts:
1. Handle any "untyped" content as string
2. If the node's type is a custom tag, shift the node's starting point in order to include it in the template

### DIFF
--- a/openformats/formats/yaml/utils.py
+++ b/openformats/formats/yaml/utils.py
@@ -67,6 +67,19 @@ class TxYamlLoader(yaml.SafeLoader):
             )
         return super(TxYamlLoader, self).compose_node(parent, index)
 
+    def _is_custom_tag(self, tag):
+        """
+        Check whether a value is tagged with a custom type.
+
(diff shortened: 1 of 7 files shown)


SofiaMargariti:
@rigaspapas there is a problem here, which I am not sure if and how we can solve it. If you compile the file without the template, custom tags are not preserved in the translation file.

rigaspapas:
This won't move forward. An alternative implementation exists in `yaml-custom-tags-alt` branch. A new PR will be opened.

PR #103, opened 2018-05-09
Improve YAML parser to avoid memory-eating behaviour

rigaspapas:
Checklist (for the reviewer)
----------------------------

* [N/A] Change is covered by unit-tests
* [x] Code is well documented
* [N/A] Errors are handled properly
* [N/A] All (conceivable) edge-cases are handled

Problem
-------

YAML parser preserved the styles of the previous elements of a list, causing huge lists to crash since the parsing of an element required a huge array with unusable information.

Steps to reproduce
------------------

Upload a YAML file with a huge lists:
```
master_key:
  - key1: val
  - key2: val
...
  - key100000: val
```

Solution
--------

Avoid unused copies of the styles array. Rewrite the code that handles style inheritance.

Example run / Screenshots
-------------------------

Performance on live data
------------------------

### DIFF
--- a/openformats/formats/yaml/yaml.py
+++ b/openformats/formats/yaml/yaml.py
@@ -232,30 +232,32 @@ class YamlHandler(Handler):
         if isinstance(yaml_data, dict):
             for key, node in yaml_data.items():
                 node_key = self._get_key_for_node(key, parent_key)
-                style = copy.copy(parent_style)
+                # Copy style for each node to avoid getting affected from the
+                # previous loops
+                node_style = copy.copy(parent_style)
                 if isinstance(node.value, dict):
                     if self.is_pluralized(node.value):
                         parsed_data.append(
                             self._parse_pluralized_leaf_node(
                                 node, node_key,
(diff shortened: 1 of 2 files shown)


openformats/formats/yaml/yaml.py
```diff
@@ -37,6 +37,9 @@ class YamlHandler(Handler):
 
     language_code = None
 
+    # Keep the parsed nodes in this global variables while parsing a file
+    parsed_data = None
+
```
kouk: technically this is a class variable, not a global variable. More-over, by assigning something to `self.parsed_data` your are converting it from a class variable to an instance variable. So my suggestions are 1) remove this line entirely and 2) rename `self.parsed_data` to `self._parsed_data` so that it's private.

rigaspapas: @kouk you are right. I need to re-phrase. By global I mean shared within the class. How about: "Helper to store the processed data while parsing a file" as a comment. I agree to rest of the proposed changes.

PR #104, opened 2018-05-29
Preserve string styles in YAML i18n plural rules

rigaspapas:
Make the i18n YAML parser to preserve the style of each plural rule. This will preserve the quotations of the source language for each rule. If the translation contains a rule that is not present in the source
language, we use the style of the singular rule.

Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------

Plural rules were always printed without quotes in the translation files, even when the source file used quotes for them.

Steps to reproduce
------------------

Upload and translate the following i18n YAML file on Transifex:
```
en:
  test:
    one: "this has quotes"
    many: 'this is single quoted'
```
When you download the translated one (no matter the mode), the quotes are missing.

Solution
--------

Keep the styles of each plural rule as serialized dictionary. When compiling plural rules, we parse and use this dictionary to retrieve the style of each rule. 

Example run / Screenshots
-------------------------

Performance on live data
------------------------

### DIFF
--- a/openformats/formats/yaml/yaml_i18n.py
+++ b/openformats/formats/yaml/yaml_i18n.py
@@ -1,4 +1,6 @@
 from __future__ import absolute_import
+import json
+import re
 
 from openformats.exceptions import ParseError
 from openformats.formats.yaml import YamlHandler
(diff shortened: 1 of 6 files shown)


dbendilas:
For the record, the commit was directly merged.

PR #112, opened 2018-09-03
Chromel18n format

Diontsoumas:
Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem
-------
The old `ChromeI18n` handler was extending the old `Json` one, which was not working with templates. In order to support this, we need to extend the new `KEYVALUEJSON` format and move the `ChromeI18n` handler to the Openformats repository.

Steps to reproduce
------------------
Since this is a new format, there is no way to compare its parsing and compiling functionalities outside of TXC. You should run the testbed to see the new format in action.

Solution
--------
For the new `ChromeI18n` handler we extend the functionality of the `Json` one, overwriting some of it when needed due to the differences of the two formats.

Example run / Screenshots
-------------------------
You can try the following `.json` file for your tests.

```
{
  "unlock": {
    "message": "Unlock",
    "description": "Unlock"
  },
  "credentials_in_db": {
    "message": "$AMOUNT$ credentials in the database",
    "description": "Amount of the credentials in the database",
    "placeholders": {
      "amount": {
        "content": "$1",
        "example": "2"
      }
    }
  }
}
```

Performance on live data
------------------------


openformats/formats/json.py
```diff
@@ -27,14 +27,15 @@ class JsonHandler(Handler):
...
+    json_dict = False
 
     PLURAL_ARG = 'plural'
     PLURAL_KEYS_STR = ' '.join(Handler._RULES_ATOI.keys())
 
     def parse(self, content, **kwargs):
         # Do a first pass using the `json` module to ensure content is valid
         try:
-            json.loads(content)
+            self.json_dict = json.loads(content)
```
Diontsoumas: This variable here worries me a bit performance wise. The `Chrome` format assumes some values have the following structure:
"""
{
'message': 'this-is-a-message'
'description:' 'this-is-the-description-of-the-message-above'
}
"""
where we set `description` to be the developer comment to the `message` value. This means that we want access to the `description` attribute when we create the `Openstring` object, but because the original `Json` handler does not share the same logic, we don't have anywhere saved this information.

My worry is that for large files, the `json_dict` variable will consume a bit of memory. Is this really an issue?

Is it worthing to overwrite the `_parse` function and assign te variable only for the Chrome handler and not for both formats?

openformats/formats/json.py
```diff
@@ -83,8 +84,8 @@ def _extract(self, parsed, nest=None):
                         openstring = self._get_regular_string(
                             key, value, value_position
                         )
-
-                    self.stringset.append(openstring)
+                    if openstring:
+                        self.stringset.append(openstring)
```
Diontsoumas: For `Chrome` not all strings will be parsed, only the ones with the `message` value.

openformats/formats/json.py
```diff
@@ -448,11 +449,12 @@ def _insert_item(self, value, value_position, is_real_stringset):
                     value, value_position, string, is_real_stringset
                 )
 
-            # Anything else: just remove the current section
             else:
-                self._copy_until_and_remove_section(
-                    value_position + len(value) + 1
-                )
+                if self._remove_string_not_in_stringset():
```
Diontsoumas: `JSON` handler will remove everything that does not exist in the stringset. For `Chrome` we want to add those strings in the template.

openformats/formats/json.py
```diff
@@ -722,3 +736,92 @@ def _unescape_generator(string):
...
+        """Return the 'description' child for a given key
+
+        :param str key: the key to search against
+        :raises ValueError: if the key has not a description child
+        :return: the description string
+        :rtype: str
+        """
+        key_split = key.split('.')
+        try:
+            return self.json_dict[key_split[0]]['description']
```
Diontsoumas: Here we take the `description` value from the `json` dictionary which we hold in memory

openformats/tests/formats/chrome_i18n/files/1_en.json
```diff
@@ -0,0 +1,16 @@
...
+    "message": "$AMOUNT$ credentials in the database",
+    "description": "Amount of the credentials in the database",
+    "placeholders": {
+      "amount": {
+        "content": "$1",
+        "example": "2"
+      }
+    }
+  }
+}
```
rigaspapas: We need to cover more cases here. I have the following scenarios in mind:
1. An entry with `@`s in the key
2. An entry with no description
3. An entry with some random keys (no message or description)
4. An entry with unicode characters
5. An entry with an integer value
Let's try to break it!

openformats/formats/json.py
rigaspapas: Sounds good. But I think that this behaviour should be reflected in the tests

openformats/formats/json.py
rigaspapas: As I mention below, we can save the file only in the Chrome parser. We only need to store the `description` values for each key. So an optimization would be to store a mapping like:
```
{"key_name": "description", "key2": ..., "keyN": ...}
```
and not save the whole JSON file. It will require some extra parsing though. Let's not worry about the performance yet. Keeping a dictionary in the memory is not a big deal for files that are a few Megabytes in size.

openformats/formats/json.py
```diff
@@ -722,3 +736,92 @@ def _unescape_generator(string):
...
+
+        :param str template: the template string
+        :param stringset: generator that holds a list of OpenString objects
+        :return: the compiled template
+        :rtype: str
+        """
+        stringset = list(stringset)
+        return self._replace_translations(template, stringset, True)
+
+    def _remove_string_not_in_stringset(self):
```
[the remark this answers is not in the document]
Diontsoumas: I thought about it but preferred to take advantage of the docstring to explain the reason behind its functionality. Do you think its better to convert it to an attribute instead?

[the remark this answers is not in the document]
rigaspapas: If the docstring is the only reason to make it a method, then yes. You can add comments :)

openformats/tests/formats/chrome_i18n/test_chrome_i18n.py
```diff
@@ -0,0 +1,111 @@
...
+                          self.random_openstring.__dict__)
+        self.assertEquals(
+            compiled, '{"a":{"message": "%s"}}' % self.random_string
+        )
+        # Check developer comment is empty
+        self.assertEquals(stringset[0].developer_comment, "")
+        # Check attribute remove_string_not_in_stringset has the correct value
+        self.assertFalse(self.handler.remove_string_not_in_stringset)
+        # Check the JSON dict lives in memoty
+        self.assertNotEquals({}, self.handler.json_dict)
```
rigaspapas: A typo exists here: Memoty.

Why not check for actual dictionary in this case?

openformats/formats/json.py
```diff
@@ -722,3 +736,97 @@ def _unescape_generator(string):
...
+        Copy characters to the transcriber until the given position,
+        then end the current section.
+        """
+        self.transcriber.copy_until(pos)
+        self.transcriber.mark_section_end()
+        # Unlike the JSON format, do not remove the remaining section of the
+        # template
+
+    def validate_content(self, content):
+        """Validate that a given string is valid JSON format.
```
rigaspapas: * a valid Chrome i18n file

...maybe also add a `TODO` comment to say that we need to perform more validation checks in the future.

openformats/formats/json.py
```diff
@@ -722,3 +736,97 @@ def _unescape_generator(string):
...
+        and update the transcriber accordingly.
+
+        :param key: the string key
+        :param value: the translation string
+        :return: an OpenString or None
+        """
+        # We should only parse keys with a specific value ("message"). All
+        # others should be added in the template
+        if self._skip_key(key):
+            return False
```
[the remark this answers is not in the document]
Diontsoumas: @dbendilas keys are in format `parent.key` (ex. `unlock.message`, `unlock.description` etc), this is why in point 2 equality will not work out.

openformats/formats/json.py
```diff
@@ -769,10 +762,10 @@ def _get_regular_string(self, key, value, value_position):
         """
         # We should only parse keys with a specific value ("message"). All
         # others should be added in the template
-        if self._skip_key(key):
-            return False
+        if self.STRING_KEY not in key:
```
dbendilas: Should be `if self.STRING_KEY != key:`, we want equality, not partial match.

dbendilas: Sorry, I hadn't seen your comment. As discussed offline, we could use `endswith()`.

PR #121, opened 2018-11-05
Decouple ICU support from JSON parser

dbendilas:
Checklist (for the reviewer)
----------------------------

* [x] Code is well documented
* [N/A] Performs well when applied to big organizations/resources/etc from our production database
* [N/A] Proper labels have been applied

Problem
-------
The parsing of ICU message format strings was integrated directly inside the JSON parser. Now it's on a separate module, so that it can be reused elsewhere.

Solution
--------
Refactored the JSON parser, taking out the ICU-related functionality.

Steps to reproduce
------------------
N/A

### DIFF
--- a/openformats/formats/json.py
+++ b/openformats/formats/json.py
@@ -5,13 +5,13 @@ from __future__ import absolute_import
 import json
 import re
 from itertools import count
-import pyparsing
 
 from ..exceptions import ParseError
 from ..handlers import Handler
 from ..strings import OpenString
 from ..transcribers import Transcriber
 from ..utils.json import DumbJson
+from ..utils.icu import ICUHelper
 
 
 class JsonHandler(Handler):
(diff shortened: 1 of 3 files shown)


openformats/formats/json.py
```diff
@@ -116,46 +102,51 @@ def _extract(self, parsed, nest=None):
...
+        :param value: the translation string
         :return: an OpenString or None
         """
-        matches = re.match(
-            ur'\s*{\s*([A-Za-z-_\d]+)\s*,\s*([A-Za-z_]+)\s*,\s*(.*)}\s*', value
+        # First attempt to parse this as a special node, e.g. a pluralized string.
+        # If it cannot be parsed that way (returns None), parse it like
+        # a regular string.
+        icu_string = ICUParser.parse(key, value, value_position)
+        if icu_string:
```
rigaspapas: It's more Pythonic to use:
```suggestion
        if icu_string is not None:
```

dbendilas: I'd argue that this is not the case. 

Moreover, the `parse()` method is expected to return an `ICUString` instance or `None`. It will never return any other object that may be evaluated to a boolean (e.g. an empty list), so I don't se a reason to use the more verbose version here, to be honest.

PR #131, opened 2018-12-12
Add helper for plural rules placeholders

dbendilas:
Checklist (for the reviewer)
----------------------------

* [ ] Code is well documented

Problem and/or solution
-------
Add a helper function that replaces all serialized plural rule placeholders (hashes), that correspond to the source language, with those that correspond to the given target language rules.

This is useful when the template does not serialize all pluralized content for a string (e.g. "{count, plural, <hash>}"), but rather adds one hash placeholder for each rule (e.g. "{count, plural, one {<hash1>} other {<hash2>}").

In the latter case, when compiling, the existing source placeholders need to be removed completely, and the target placeholders need to appear in their place.

### DIFF
--- a/openformats/tests/util_tests/test_icu.py
+++ b/openformats/tests/util_tests/test_icu.py
@@ -4,10 +4,12 @@ import unittest
 
 from openformats.strings import OpenString
 from openformats.utils.icu import (ICUCompiler, ICUParser, ICUString,
-                                   normalize_plural_rule)
+                                   normalize_plural_rule, PLURAL_FORMAT_NUMERIC,
+                                   PLURAL_FORMAT_STRING)
 
 
 class ICUStringTestCase(unittest.TestCase):
+    """Test the ICUString class."""
 
     def test_strings_by_rule(self):
         """Make sure the strings_by_rule property works properly."""
@@ -23,6 +25,28 @@ class ICUStringTestCase(unittest.TestCase):
         self.assertDictEqual(
             icu_str.strings_by_rule,
             {0: u'0 lamparas', 1: u'1 lámpara', 5: u'{cnt} lamparas'},
+        )
+
+    def test_syntax_by_rule(self):
+        """Test the `syntax_by_rule` property."""
+        icu_str = ICUString(
+            'key',
+            [
+                ('=0', u'{0 pins}'),
+                ('=1', u'{one pin}'),
+                ('two', u'{a couple of pins}'),
+                ('other', u'{{cnt} pins}')
+            ],
+            pluralized=True,
+        )
+        self.assertDictEqual(
+            icu_str.syntax_by_rule,
+            {
+                0: PLURAL_FORMAT_NUMERIC,
+                1: PLURAL_FORMAT_NUMERIC,
+                2: PLURAL_FORMAT_STRING,
+                5: PLURAL_FORMAT_STRING,
+            },
         )
 
     def test_icustring_representation(self):
@@ -54,7 +78,7 @@ class ICUStringTestCase(unittest.TestCase):
 
 
 class ICUParserTestCase(unittest.TestCase):
-    """Test the functionality of the ICU module.
+    """Test the functionality of the ICUParser class.
 
     Most of its functionality is covered in the JSON tests, so here
     we only cover a few parts, namely:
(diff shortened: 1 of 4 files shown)


openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
 
-        :param pluralized_string: an OpenString that is pluralized
-        :param delimiter: a string to use for separating entries
+        :param OpenString pluralized_string: an OpenString that is pluralized
+        :param str delimiter: a string to use for separating entries
+        :return: a string with all plural rules and their
+            corresponding strings
+        :rtype: str
+        """
+        return self.serialize_strings(pluralized_string.string, delimiter)
```
rigaspapas: What is the purpose of this method? It just calls another

openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
+            for rule, translation in strings_by_rule.iteritems()
         ]
         return delimiter.join(plural_list)
+
+
+def update_template_for_plural_rules(content, target_plural_forms,
+                                     allow_numeric_plural_values=True):
+    """Update the given content so that it contains all
+    necessary placeholders the target language supports,
+    no more, no less.
```
rigaspapas: The "no more, no less" part can be removed. Or you can specify that this happens "independently of the source language's rules"

openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
+            line = content[offset:]
+        first_brace_pos = line.find('{')
+        # If there is no '{' found, it's definitely not
+        # a pluralized string, so move on
+        if first_brace_pos == -1:
+            # Advance the offset, so that we don't end up with an endless loop
+            offset += len(line)
+            if separator_pos > -1:
+                offset += sep_length
+            continue  # noqa
```
rigaspapas: Why the `# noqa` here?

openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
+        icu_string = parser.parse(
+            value=serialized_str,
+            key='',  # we don't care about the key
+        )
+        # If it's not formatted as an ICU pluralized string, move on
+        if not icu_string:
+            offset += len(line)
+            if separator_pos > -1:
+                offset += sep_length
+            continue  # noqa
```
rigaspapas: The 4 lines above are repeated. Should we move them in a different helper and re-use this?

openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
+        # {
+        #    1: 'a2eb8a66695435a29ee61d5df781679b_pl_0',
+        #    5: 'a2eb8a66695435a29ee61d5df781679b_pl_1',
+        # }
+        hash_str = icu_string.strings_by_rule[5]
+        hash_str = hash_str[:-1]  # remove the last char, which is the plural index
+        strings_by_rule = {
+            rule: hash_str + str(index)
+            for index, rule in enumerate(target_plural_forms)
+        }
```
rigaspapas: Since this method is big, should we move the above 3 commands to a different method?

openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
+        # e.g. 'a2eb8a66695435a29ee61d5df781679b_pl_'
+        # and create a dictionary of all hashes, one per each rule
+        # of the target language, e.g.:
+        # {
+        #    1: 'a2eb8a66695435a29ee61d5df781679b_pl_0',
+        #    5: 'a2eb8a66695435a29ee61d5df781679b_pl_1',
+        # }
+        hash_str = icu_string.strings_by_rule[5]
+        hash_str = hash_str[:-1]  # remove the last char, which is the plural index
+        strings_by_rule = {
```
rigaspapas: I think it's better to rename this as `hashes_by_rule` or `plural_placeholders`. There are no actual strings there.

openformats/utils/icu.py
```diff
@@ -384,17 +421,180 @@ def serialize_string(self, pluralized_string, delimiter=' '):
...
+    parser = ICUParser(allow_numeric_plural_values=allow_numeric_plural_values)
+    compiler = ICUCompiler()
+
+    # We'll take the content in chunks, one line at the time
+    while True:
+        separator_pos = content.find(line_separator, offset)
+
+        # If there are no more lines, stop iterating
+        if offset > 0 and separator_pos == -1:
+            break
```
rigaspapas: We could make the code more descriptive here. Something like:
```
no_more_lines = offset > 0 and separator_pos == -1
if no_more_lines:
    break
```

Should we move this to the end of the loop and change the `while` condition to something like: `while more_lines`?

openformats/tests/util_tests/test_icu.py
```diff
@@ -106,3 +132,88 @@ def test_compiler(self):
...
+            allow_numeric_plural_values=True,
+        )
+        self._test_rules(new_content, set(rules))
+
+    def _test_rules(self, new_content, rule_set):
+        icu_string = ICUParser().parse('', new_content[len('key='):])
+        self.assertSetEqual(
+            set(icu_string.strings_by_rule.keys()),
+            rule_set,
+        )
```
rigaspapas: I think that the following cases are missing from the unittests:
1. A case with `\r\n` as separator
2. A case with a line without braces
3. A case with a line with braces but without ICU contents

openformats/utils/icu.py
dbendilas: It was used by the JSON handler. You are right, it doesn't make any sense to keep it.

openformats/utils/icu.py
dbendilas: I liked the "no more, no less" part, but I like your suggestion more.

openformats/utils/icu.py
dbendilas: (this was actually replaced with a `# pragma: no cover` on the next commit, I assume you had commented on the commit diff only)

The reason this was added is that during testing, execution never reached that point. I spent 1-2 hours on this and I am pretty sure that the code was actually reached, but because of the way `continue` is handled in terms of execution, the coverage script did not understand that. I had to add the `no cover` mark, otherwise the coverage fell to a point that the PR was blocked from merging.

openformats/utils/icu.py
dbendilas: If we made such a helper, we would have to pass 3 arguments (`line` or its length, `separator_pos` and `sep_length`) in order to make it work. And we would still have to manually call `continue`. So I don't think it makes much sense, it would be just noise.

openformats/utils/icu.py
dbendilas: I tried it and I prefer the current way.

Although I would prefer to not have `while True` myself, changing the logic to `while <condition>` requires significant changes due to the the 2 `continue` blocks we have. I'm reluctant to make such as change at this point.

openformats/tests/util_tests/test_icu.py
dbendilas: Sounds good, I'll amend the tests.

dbendilas: Update: the previous implementation made assumptions about the file format (assumed key-value syntax). The functionality of the ICU methods and functions is now specific to ICU syntax only, agnostic to any specific file format. Therefore, the tests mentioned above are no longer valid.

openformats/tests/util_tests/test_icu.py
```diff
@@ -177,6 +213,31 @@ def test_with_various_strings(self):
             u'keyB={count, plural, one {hash_str_B_0} few {hash_str_B_1}'
             u' many {hash_str_B_2} other {hash_str_B_3}}\n'
             u'keyC=something\n'
+            u'keyD={no ICU brace}\n'
+            u'keyE=brace {\n'
+        )
+
+    def test_with_alternate_newline_char(self):
```
rigaspapas: I think we should use "alternative", not "alternate"

openformats/utils/icu.py
```diff
@@ -372,29 +407,133 @@ def _parse_plural_content(string):
...
+        the ICUString object. For example, if for rule 'one'
+        the guideline is to use a numeric form, '=1' will be used
+        when serializing.
+
+        In other words, what this method does is to create a serialized
+        version of placeholder hashes, following the ICU plural format,
+        for a specific set of plural rules. This works regardless
+        how many plural rules are found in `icu_string` (source language
+        string) and how many languages are found in `plural_rules`
+        (target language). The reason this works is that all placholders
```
rigaspapas: Typo: placholders

openformats/tests/util_tests/test_icu.py
```diff
@@ -96,13 +120,131 @@ def test_plural_rule_normalization(self):
...
+            [1, 2, 3, 4, 5]
+        )
+        self.assertEqual(
+            u'one {s}_0} two {s}_1} few {s}_2} many {s}_3} other {s}_4}'.replace(
+                u'{s}', u'{ac76ac7a27_pl',
+            ),
+            serialized,
+        )
+
+    def test_with_less_target_languages(self):
```
rigaspapas: I think that this test should be named `test_with_less_target_plural_rules`, not `languages`. Also, the plural rules are equal in quantity but different. Should we add some more tests on this?

rigaspapas: I think that this test should be named `test_with_less_target_plural_rules`, not `languages`. Also, the plural rules are equal in quantity but different. Should we add some more tests on this?

openformats/tests/util_tests/test_icu.py
dbendilas: Test was removed.

PR #136, opened 2019-02-12
Fix various YAML-related bugs

dbendilas:
Problem and/or solution
-----------------------
There were 3 different scenarios in which the compilation of the
YAML front matter section was failing. All 3 had to do with wrapping
the string in quotes, and all were fixed.

- Wrap in double quotes if the string starts with `[` but does not end
  in `]`
- Wrap in double quotes if the string starts with `@`
- Wrap in single quotes if the string starts with double quotes but
  does not end in double quotes

How to test
-----------
Unit tests.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/github_markdown_v2.py
+++ b/openformats/formats/github_markdown_v2.py
@@ -19,22 +19,16 @@ class GithubMarkdownHandlerV2(OrderedCompilerMixin, Handler):
 
     BACKSLASH = u'\\'
     DOUBLE_QUOTES = u'"'
+    SINGLE_QUOTE = u"'"
     NEWLINE = u'\n'
     COLON = u':'
     ASTERISK = u'*'
     AMPERSAND = u'&'
     DASH = u'-'
     HASHTAG = u'#'
-
-    def _should_wrap_in_quotes(self, tr_string):
-        return any([
-            self.NEWLINE in tr_string[:-1],
-            self.COLON in tr_string,
-            self.HASHTAG in tr_string,
-            tr_string.lstrip().startswith(self.ASTERISK),
-            tr_string.lstrip().startswith(self.AMPERSAND),
-            tr_string.lstrip().startswith(self.DASH),
-        ])
+    AT_SIGN = u'@'  # reserved character in the YAML spec
+    BRACKET_LEFT = u'['
+    BRACKET_RIGHT = u']'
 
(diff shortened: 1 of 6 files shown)


openformats/formats/github_markdown_v2.py
```diff
@@ -135,3 +112,129 @@ def parse(self, content, **kwargs):
...
+
+    def _should_wrap_in_quotes(self, string):
+        """Check if the given string should be wrapped in quotes.
+
+        In order to decide if wrapping is necessary, it takes into account
+        various parameters, such as what character the string starts and ends with,
+        whether or not it contains special characters, etc.
+
+        :param unicode string: the string to check
+        :return: a tuple that should if wrapping is needed and the character that
```
Diontsoumas: This return sentence seems incomplete. The tuple should what?

openformats/formats/github_markdown_v2.py
```diff
@@ -135,3 +112,129 @@ def parse(self, content, **kwargs):
...
+        :param str quote_char: the character to use for wrapping
+        :return: the new string, wrapped in quotes if needed
+        :rtype: unicode
+        """
+        if quote_char == u'"':
+            string = string.replace(quote_char, self.BACKSLASH + quote_char)
+        elif quote_char == u"'":
+            string = string.replace(quote_char, quote_char * 2)
+        else:
+            raise ValueError(
```
Diontsoumas: Please add `:raise:` to the docstring

openformats/tests/formats/github_markdown_v2/test_github_markdown.py
```diff
@@ -29,3 +31,104 @@ def test_parse(self):
         content_with_tab = self.handler.parse(content=u"# foo	bar")
         content_with_spaces = self.handler.parse(content=u"# foo    bar")
         self.assertEqual(content_with_tab[0], content_with_spaces[0])
+
+
+class GithubMarkdownV2CustomTestCase(unittest.TestCase):
+
```
dbendilas: @dbendilas note to self: add a docstring here.

openformats/formats/github_markdown_v2.py
dbendilas: Good catch!

PR #143, opened 2019-05-03
Po download obsoleted strings fix

kostasgeo:
Problem and/or solution
-----------------------
When uploading a PO file in Resources, if obsoleted strings (marked with `#~`) are part of the file, due to the separate handling of obsolete and non obsolete entries in polib, the template generated at PO file handler's parsing has different order than the strings in the original file. Thus, the information saved in the database and as a consequence in the exported file, misses any non obsoleted strings that follow the obsoleted ones.

As a solution, we by pass the polib.POFile 's __unicode__() method by not generating the template using six.text_type(po_file) and instead provide a method in PoHandler that generates the template without changing the order when obsoleted string are present.

How to test
-----------

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/po.py
+++ b/openformats/formats/po.py
@@ -45,7 +45,33 @@ class PoHandler(Handler):
             openstring = self._handle_entry(entry)
             if openstring is not None:
                 stringset.append(openstring)
-        return six.text_type(self.new_po), stringset
+
+        po_str = self._po_file_to_str(self.new_po)
+        return po_str, stringset
+
+    def _po_file_to_str(self, po_file):
+        """
+        Generate the template for the PO file.
+
+        Use this method instead of six.text_type(po_file)
+        in order to avoid the separate handling of obsoleted strings
+        by the PO file handler of polib in its __unicode__ method,
+        since this affects the order of strings in the tempate and
+        as a consequence, strings after obsoleted ones are missing
+        in the compiled file.
+        """
+        result = []
+        headers = po_file.header.split('\n')
+        for header in headers:
+            if header[:1] in [',', ':']:
+                result.append('#%s' % header)
+            else:
+                result.append('# %s' % header)
+        result.append(six.text_type(po_file.metadata_as_entry()))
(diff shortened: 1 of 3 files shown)


openformats/formats/po.py
```diff
@@ -45,7 +45,23 @@ def parse(self, content, is_source=False):
             openstring = self._handle_entry(entry)
             if openstring is not None:
                 stringset.append(openstring)
-        return six.text_type(self.new_po), stringset
+
+        po_str = self._po_file_to_str(self.new_po)
+        return po_str, stringset
+
+    def _po_file_to_str(self, po_file):
+        ret, headers = '', po_file.header.split('\n')
```
dbendilas: I just had a glance out of curiosity; please add a docstring to explain what this method does.

kostasgeo: @dbendilas  thank you, I just added it.

openformats/formats/po.py
```diff
@@ -45,7 +45,23 @@ def parse(self, content, is_source=False):
             openstring = self._handle_entry(entry)
             if openstring is not None:
                 stringset.append(openstring)
-        return six.text_type(self.new_po), stringset
+
+        po_str = self._po_file_to_str(self.new_po)
+        return po_str, stringset
+
+    def _po_file_to_str(self, po_file):
```
igavriil: Generally, given a string is preferably (performance wise) to create a list of strings and use `join` instead of appending with `+`. With this in mind i would change this method to operate on top of a list of strings:
```
result = []
for header in headers:
    if header[:1] in [',', ':']:
        result.append(six.text_type('#%s' % header))
    else:
        result.append(six.text_type('# %s' % header))

result.append(po_file.metadata_as_entry())

for entry in po_file:
    result.append(six.text_type(entry))

return six.text_type('\n'.join(result))
```

igavriil:
Looks good! Can you please add a more descriptive commit message to describe both the issue and the solution given?

PR #164, opened 2019-07-23
Escape control characters on compile

dbendilas:
Problem and/or solution
-----------------------
Escape special control characters and other non-printable characters. Use Unicode escaped format as a replacement when compiling. For example, the backspace control character gets escaped to `\u0008`.

See: https://yaml.org/spec/1.2/spec.html#id2770814 for the spec

How to test
-----------
Unit tests.

Also:

(1) upload the following file to Transifex:
```
---
something:
  key: This is something

---
```
(2) Translate using `ab` as the translation text (there is a backspace char between `a` and `b`, so copy + paste from here)

(3) Download for use. The downloaded file should look like this
```
---
something:
  key: a\u0008b

---
```

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/github_markdown_v2.py
+++ b/openformats/formats/github_markdown_v2.py
@@ -5,6 +5,8 @@ import re
 import six
 
 from mistune import Markdown
+from yaml.reader import Reader
+
 from openformats.formats.github_markdown import TxBlockLexer, string_handler
 from openformats.formats.yaml import YamlHandler
 from openformats.utils.compat import ensure_unicode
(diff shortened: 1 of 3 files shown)


openformats/formats/github_markdown_v2.py
```diff
@@ -39,15 +41,16 @@ def compile(self, template, stringset, **kwargs):
         transcriber = Transcriber(template)
         template = transcriber.source
 
-        for string in stringset:
-            tr_string = string.string
-            if self._is_yaml_string(string):
-                tr_string = self._transform_yaml_string(string)
+        for openstring in stringset:
```
dbendilas: Note to reviewer: renamed `string` to `openstring` for better clarity.

openformats/formats/github_markdown_v2.py
```diff
@@ -137,6 +140,28 @@ def _is_yaml_string(self, string):
...
+        """
+        # Check each plural rule of the string
+        for rule, string in six.iteritems(openstring.strings):
+            chars = []
+            # Go through each character
+            # If a control character is found (e.g. backspace)
+            # escape it to a unicode format, e.g. \u0008
+            for x in string:
+                if Reader.NON_PRINTABLE.match(x):
+                    chars.append('\\u{:04d}'.format(ord(x)))
```
deathbird: I think (since I don't have much experience with unicode and code-points) the above conversion of the non-printable char to a python unicode string is wrong: I tested with the DEL character which is \u007f but got back \u0127. From ord() doc saw that it returns an integer. I think you should convert it to hex since \u escape sequence is followed by four hex digits.

openformats/strings.py
```diff
@@ -110,6 +110,10 @@ def string(self):
         else:
             return self._strings[5]
 
+    @property
+    def strings(self):
+        return self._strings
```
deathbird: I see that _strings field is populated differently when the OpenString is pluralized. Can we test against a pluralized resource?

PR #167, opened 2019-09-17
Unescape non-printable chars in markdown YAML headers

dbendilas:
Problem and/or solution
-----------------------
YAML defines a set of characters that are considered to be invalid. It accepts only printable characters: https://yaml.org/spec/1.2/spec.html#id2770814

Specifically, it says that:
> On input, a YAML processor must accept all Unicode characters except those explicitly excluded above.
> On output, a YAML processor must only produce acceptable characters. Any excluded characters must be presented using escape sequences. In addition, any allowed characters known to be non-printable should also be escaped.

The GithubMarkdown format has a YAML header, which must too conform to these specs. The latter was handled in #164. The former is handled by this PR. Note that both aspects are only handled for the YAML part of the GithubMarkdown format, not the YAML format. 

More specifically, when parsing the YAML header, any escaped non-printable character (i.e. a string with length > 1 in Python), such as `\x03` and `\u0004` is converted to the equivalent character (one-character string in Python).

How to test
-----------
1. Unit tests.

2. Use the following content on the testbed:
```yaml
---
test:
  should_be_escaped: \x01\x02\x03\x04\x05\x06\x07\x08\x10\x11\x12\x1e\x1F\x7f
  should_be_kept: \x85\x00\x09\xA9

---

A line is needed otherwise all strings in the YAML header are parsed as one (no clue why).

```
3. Upload a Github Markdown resource with the above content and see how it's parsed (look at the Editor).

#### NOTE
When compiling a characters with the `\xNN` format, the compiler cannot know that they were previously formatted like this and thus it converts them to the equivalent `\uNNNN` notation. For example, `\x08` will become `\u0008` when compiled.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/github_markdown_v2.py
+++ b/openformats/formats/github_markdown_v2.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, unicode_literals
 
 import re
 
@@ -23,18 +23,23 @@ class GithubMarkdownHandlerV2(OrderedCompilerMixin, Handler):
     extension = "md"
     EXTRACTS_RAW = False
 
-    BACKSLASH = u'\\'
-    DOUBLE_QUOTES = u'"'
+    BACKSLASH = '\\'
+    DOUBLE_QUOTES = '"'
     SINGLE_QUOTE = u"'"
-    NEWLINE = u'\n'
-    COLON = u':'
-    ASTERISK = u'*'
-    AMPERSAND = u'&'
-    DASH = u'-'
-    HASHTAG = u'#'
-    AT_SIGN = u'@'  # reserved character in the YAML spec
-    BRACKET_LEFT = u'['
-    BRACKET_RIGHT = u']'
+    NEWLINE = '\n'
+    COLON = ':'
+    ASTERISK = '*'
+    AMPERSAND = '&'
+    DASH = '-'
+    HASHTAG = '#'
+    AT_SIGN = '@'  # reserved character in the YAML spec
+    BRACKET_LEFT = '['
+    BRACKET_RIGHT = ']'
+
+    # A string that looks like '\u0008'
+    ESCAPED_UNICODE = re.compile(r'\\u[a-fA-F0-9]{4}')
+    # ... or '\x2D'
+    ESCAPED_UNICODE_HEX = re.compile(r'\\x[a-fA-F0-9]{2}')
 
     def compile(self, template, stringset, **kwargs):
         # assume stringset is ordered within the template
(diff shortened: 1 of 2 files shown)


openformats/formats/github_markdown_v2.py
```diff
@@ -36,6 +36,11 @@ class GithubMarkdownHandlerV2(OrderedCompilerMixin, Handler):
     BRACKET_LEFT = u'['
     BRACKET_RIGHT = u']'
 
+    # A string that looks like '\u0008'
+    ESCAPED_UNICODE = re.compile(u'\\\\u[a-fA-F0-9]{4}')
+    # ... or '\x2D'
+    ESCAPED_UNICODE_HEX = re.compile(u'\\\\x[a-fA-F0-9]{2}')
```
kbairak: These `\\\\` thingies are scary. How about we set `from __future__ import unicode_literals` on top and use `r''` strings?

openformats/formats/github_markdown_v2.py
```diff
@@ -141,27 +150,81 @@ def _is_yaml_string(self, string):
...
+
+        :param OpenString openstring: the string object to check and update
+        """
+        # Check each plural rule of the string
+        for rule, string in six.iteritems(openstring.strings):
+            # Check \u0000 notation
+            matches = GithubMarkdownHandlerV2.ESCAPED_UNICODE.finditer(string)
+            all_matches = []
+            for match in matches:
+                all_matches.append(match)
```
kbairak: `all_matches = list(GithubMarkdownHandlerV2.ESCAPED_UNICODE.finditer(string))` ?

openformats/formats/github_markdown_v2.py
```diff
@@ -141,27 +150,81 @@ def _is_yaml_string(self, string):
...
+
+            # Check \x00 notation
+            matches = GithubMarkdownHandlerV2.ESCAPED_UNICODE_HEX.finditer(string)
+            all_matches = []
+            for match in matches:
+                all_matches.append(match)
+
+            for match in reversed(all_matches):
+                escaped = match.group()
+                print(repr(escaped))
```
kbairak: Don't forget to remove this `print`

openformats/formats/github_markdown_v2.py
dbendilas: Unfortunately it doesn't work, that's why I had to do this workaround.

openformats/formats/github_markdown_v2.py
dbendilas: Oops, thanks.

openformats/formats/github_markdown_v2.py
dbendilas: I didn't do that because I'd have to change all `u''` occurrences in that module. But I agree, `\\\\` looks awful. I've made the change now.

openformats/formats/github_markdown_v2.py
```diff
@@ -141,27 +150,77 @@ def _is_yaml_string(self, string):
...
             # escape it to a unicode format, e.g. \u007f
-            for x in string:
-                if Reader.NON_PRINTABLE.match(x):
-                    chars.append('\\u{:04x}'.format(ord(x)))
+            for char in string:
+                if Reader.NON_PRINTABLE.match(char):
+                    chars.append('\\u{:04x}'.format(ord(char)))
                 else:
-                    chars.append(x)
+                    chars.append(char)
```
dbendilas: I just replaced `x` with `char` for better clarity.

openformats/formats/github_markdown_v2.py
dbendilas: > Unfortunately it doesn't work, that's why I had to do this workaround.

I take that back. I used to get `[]` all the time, but maybe I was doing something wrong with the tests. Let's manual test it again to make sure @kbairak.

PR #168, opened 2019-09-19
Reimplementation of InDesign handler

kbairak:
Problem and/or solution
-----------------------

It turns out that using UCFLib can damage the idml archive. The
indications we have of this is that files compiled by this handler:

1. Show up as "corrupted" in InDesign
2. Have messed up permissions when extracted by archive managers (tested
   on Linux and MacOS operating systems)

The new implementation operates on the archives directly. We access
'designmap.xml' to gather the story IDs, then gather all the story files
and extract the content their <Content> tags as templates. Finally we
create a new archive identical in every way possible to the old one but
with the stories replaced by the templates. A very similar process is
performed during compilation.

It turns out that trying to retain as much of the structure and content
of the resulting archive fixes our issues.

How to test
-----------

Theoretically, the output of parsing and compiling, at least in terms of
strings and individual story templates, should be the same as before,
thus the same tests should pass.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/indesign.py
+++ b/openformats/formats/indesign.py
@@ -203,6 +203,13 @@ class InDesignHandler(Handler):
 
     @staticmethod
     def _make_zipfile_copy(zipfile_in, data):
+        """ Copy most of 'zipfile_in' to a new zip file. If a filename in
+            'zipfile_in' exists as a key in 'data', then instead of copying
+            that file from 'zipfile_in', write the relevant value from 'data'.
+
+            Return the byte contents of the new zipfile.
+        """
+
         file_out = io.BytesIO()
         zipfile_out = zipfile.ZipFile(file_out, 'w')
         for filename in zipfile_in.namelist():
--- a/openformats/formats/indesign.py
+++ b/openformats/formats/indesign.py
@@ -1,77 +1,158 @@
 from __future__ import absolute_import
 
 import io
+import re
+import unicodedata
 import zipfile
 from itertools import count
 
 import six
-
 from lxml import etree
+
 from openformats.handlers import Handler
 from openformats.strings import OpenString
 from openformats.transcribers import Transcriber
-from openformats.utils.xml import NewDumbXml as DumbXml
+from openformats.utils.compat import ensure_unicode
 
 
 class InDesignHandler(Handler):
+    """ A handler class that parses and compiles .idml files that are created
+        in Adobe's InDesign.
+
+        IDML files contain multiple XML fragments that can be parsed to extract
+        strings from.
+    """
+
     name = "InDesign"
     extension = "idml"
     SPECIFIER = None
     PROCESSES_BINARY = True
     EXTRACTS_RAW = False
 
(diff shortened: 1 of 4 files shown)


openformats/formats/indesign.py
```diff
@@ -122,103 +121,92 @@ def _can_skip_content(self, string):
...
+                transcriber.skip(len(self.string.template_replacement))
+                self.string = next(self.stringset, None)
+
+        transcriber.copy_to_end()
+        compiled = transcriber.get_destination()
+        compiled = hash_regex.sub(u'', compiled)
+        return compiled
+
+    @staticmethod
+    def _get_story_filenames(z_in):
```
rigaspapas: Same here. I prefer to rename `z_in` to something like `zip_file`

openformats/formats/indesign.py
```diff
@@ -122,103 +121,92 @@ def _can_skip_content(self, string):
...
+        """ Find the filenames of all Story fragmens by intersecting the list
+            of keys of the 'StoryList' attribute of the top node in
+            'designmap.xml' with the filenames within the archive that refer to
+            Story fragments.
```
rigaspapas: It is not clear for me what is the goal of this method based on the docstring. Is it to remove the stories that are not referenced in `designmap.xml`? Is it the opposite? Can you explicitly mention it?

openformats/formats/indesign.py
```diff
@@ -122,103 +121,92 @@ def _can_skip_content(self, string):
...
+              much of the structure and content as possible, but replacing the
+              story fragments with the compiled ones
+            * Return the compiled archive
+        """
 
-        out = io.BytesIO()
-        idml.save(out)
-        return out.getvalue()
+        self.stringset, compiled_dict = iter(stringset), {}
+        self.string = next(self.stringset)
```
rigaspapas: So, in order for _compile_story to work, we need to initialize the `self.string` to the first item of the `self.stringset` list. Is this correct? Why not make `_compile_story` iterate through `self.stringset` and depend on an external initialization?

openformats/formats/indesign.py
kbairak: I will rewrite this hoping it's better next time

openformats/formats/indesign.py
kbairak: This comes up fairly regularly in openformats. The reason is that you are trying to iterate over a stringset and the "sections" of a template at the same time. Also, some sections do not correspond to strings because they may have been removed from the stringset because the string is untranslated. After having written a few handlers, I have concluded that `stringset = iter(stringset); string = next(stringset, None)` is the most convenient approach.

If you look at the old implementation, it did something like:

```python
for section in sections:
    string = stringset.pop(0)
    if string_matches_section(string, section):
        add_string_to_section(string, section)
    else:
        stringset.insert(string, 0)
```

I think that having the next string at the ready, adding checks if it's `None` (because we ran out of strings before we finished all the sections) and fetching the next string if we use the current string when compiling is the simplest to read and fastest approach.

PR #170, opened 2019-09-23
Escape `&` when compiling IDML (InDesign) files

kbairak:
Problem and/or solution
-----------------------

"Lonely" ampersands (`&`) will make the resulting XML story fragments
invalid and subsequently InDesign will not be able to import the file
and it will appear corrupt. This commit forces the escaping of lonely
`&` to the proper XML escape sequence: `&amp;`.

If, however, a valid escape sequence appears in the string (eg `&lt;` or
`&#x0a1f;`), we will assume that the translator knew what they were
doing and we will not escape the sequence's first character.

So, for example, the following escapes will occur:

    "hello world"         -> "hello world"
    "hello &world"        -> "hello &amp;world"
    "hello &amp;world"    -> "hello &amp;world"
    "hello &lt;world"     -> "hello &lt;world"
    "hello &#x0a1f;world" -> "hello &#x0a1f;world"
    "&&#x05af;&&"         -> "&amp;&#x05af;&amp;&amp;"

How to test
-----------

It is a bit hard to test without having InDesign installed, but if you read and run the new unit-test you should be comfortable with the result.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/indesign.py
+++ b/openformats/formats/indesign.py
@@ -212,7 +212,7 @@ class InDesignHandler(Handler):
                 break
             else:
                 transcriber.copy_until(hash_position)
-                transcriber.add(current_string.string)
+                transcriber.add(self._escape_amps(current_string.string))
                 transcriber.skip(len(current_string.template_replacement))
 
         # Update the XML file to contain the template strings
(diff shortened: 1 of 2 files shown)


openformats/formats/indesign.py
```diff
@@ -222,3 +222,39 @@ def _compile_story(self, story_content):
         # them with an empty string
         compiled_story = hash_regex.sub(u'', compiled_story)
         return compiled_story
+
+    @staticmethod
+    def _escape_amps(string):
+        """ Escapes "lonely" `&` (ampersands). If a valid XML escape sequence
```
dbendilas: According to the TEM, this should be
```
"""Escape "lonely" `&` (ampersands). 

If a valid XML escape sequence...
"""
```
i.e. ~escapes~ escape, and a new line for the detailed description.

PR #183, opened 2020-06-03
Throw ParseError on duplicate YAML keys

ThemisB:
Problem
------------
Duplicate keys are not allowed, according to the YAML specification. Under the hood, we use the PYYAML library. For the time being, PYYAML parser accepts duplicate yaml keys and this is an issue that [has been reported by the Open Source Community](https://github.com/yaml/pyyaml/issues/165).

Solution
-----------
We post process the generated pairs to verify that we do not have duplicate keys.

How to test
-----------
Run `python setup.py test`. A new test `test_parse_duplicate_keys (openformats.tests.formats.yaml.test_yaml.YamlTestCase)` has been added.
 
Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/yaml/utils.py
+++ b/openformats/formats/yaml/utils.py
@@ -95,6 +95,8 @@ class TxYamlLoader(yaml.SafeLoader):
         Override `yaml.SafeLoader.construct_mapping` to return for each item
         of the mapping a tuple of the form `(key, (value, start, end, style,
         tag))` instead of the default which is `(key, value)`.
+        :raise ParseError: if node is not a MappingNode
+            or duplicate keys are found.
         """
         if not isinstance(node, yaml.MappingNode):
             raise ParseError(
@@ -161,6 +163,26 @@ class TxYamlLoader(yaml.SafeLoader):
 
             value = Node(value, start, end, style, tag)
             pairs.append((key, value))
+
+        # If there are duplicate keys, throw an exception
+        pair_keys = [pair[0] for pair in pairs]
+        seen = set()
+        duplicates = set()
+        seen_add = seen.add
+        duplicate_add = duplicates.add
+        for x in pair_keys:
+            if x not in seen:
+                seen_add(x)
+            else:
+                duplicates_add(x)
+
+        if len(duplicates):
+            duplicates_list = list(duplicates)
+            error_duplicate_keys = ','.join(key for key in duplicates_list)
+            raise ParseError(
+                "Duplicate keys found ({})".format(error_duplicate_keys)
+            )
+
         return pairs
 
     def construct_sequence(self, node, deep=True):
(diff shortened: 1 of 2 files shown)


openformats/formats/yaml/utils.py
```diff
@@ -161,6 +161,23 @@ def construct_mapping(self, node, deep=True):
...
             pairs.append((key, value))
+
+        pair_keys = [pair[0] for pair in pairs]
+        seen = set()
+        duplicates = set()
+        for x in pair_keys:
+            if x not in seen:
+                seen.add(x)
+            else:
+                duplicates.add(x)
```
dbendilas, line 172: This is _almost_ the fastest implementation you can have. You can make the performance 30% faster by storing the 2 `add()` methods, to avoid method lookup inside the loop:

```python
def themis_opt(pair_keys):
    seen = set()
    duplicates = set()
    seen_add = seen.add
    duplicate_add = duplicates.add
    for x in pair_keys:
        if x not in seen:
            seen_add(x)
        else:
            dup_add(x)
    return list(duplicates)
```

See this for more: https://stackoverflow.com/a/9836685/809414

dbendilas: My benchmarks:

```python
def tt(f, l):
    t = time()
    for x in range(100000):
    f(l)
    print('Elapsed: {}'.format(time() - t))

ll = [1,2,3,2,1,5,6,5,5,5] * 100
ll = [str(x) for x in ll]
```

```python
tt(themis, ll)
# Elapsed: 8.05499911308

tt(themis_opt, ll)
# Elapsed: 5.62615799904
```

openformats/formats/yaml/utils.py
```diff
@@ -161,6 +161,23 @@ def construct_mapping(self, node, deep=True):
...
+
+        if (len(duplicates)):
```
dbendilas, line 174: External parentheses are redundant.

openformats/formats/yaml/utils.py
```diff
@@ -161,6 +161,23 @@ def construct_mapping(self, node, deep=True):
 
             value = Node(value, start, end, style, tag)
             pairs.append((key, value))
+
+        pair_keys = [pair[0] for pair in pairs]
```
dbendilas, line 165: The code that checks for duplicates spans many lines of code, so I suggest a comment in the beginning, to make it easier for the user to read. Something like:

```python
# If there are duplicate keys, throw an exception
```

dbendilas: Also, it would be nice to update the docstring of `construct_mapping` with something like:
```python
:raise ParseError: if duplicate keys are found
```

openformats/tests/formats/yaml/test_yaml.py
```diff
@@ -146,3 +146,13 @@ def test_openstring_attributes(self):
...
+
+    def test_parse_duplicate_keys(self):
+        filepath = path.join(
+            self.TESTFILE_BASE, "1_en_with_duplicate_keys.yml"
+        )
+        with open(filepath, "r", encoding='utf-8') as myfile:
+            content = myfile.read()
+
+        with self.assertRaises(ParseError):
+            self.handler.parse(content)
```
dbendilas, line 158: It's OK to use a file for this test, but I think it's better to avoid using one, as test files are typically used for full-blown examples and in almost all cases follow a certain pattern (`1_en.yml`, `1_el.yml`, `1_en_exported.yml` etc) for all file formats.

Since the content you want to test is small, I would add it inline in the code instead of a file.

PR #194, opened 2020-09-29
Remove trailing newline in pluralized strings

sergafts:
When YAML files are parsed, pluralized strings span across multiple
lines. In this case, we shouldn't add a new line after each string
during compilation since the template already includes the required
newline.

Problem and/or solution
-----------------------

How to test
-----------

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/tests/formats/yamlinternationalization/files/1_en_exported.yml
+++ b/openformats/tests/formats/yamlinternationalization/files/1_en_exported.yml
@@ -3,7 +3,6 @@ en:
   pluralized_string:
     one: One ȧɠǿ
     other: Other ȧɠǿ
-
   another normal string: normal string
 
   empty_pluralized_string:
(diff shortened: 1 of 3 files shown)


openformats/tests/formats/yamlinternationalization/files/1_en_exported.yml
```diff
@@ -15,11 +14,10 @@ en:
     one: 'this is a test'
     other: "this is a double quote test"
 
-
   pluralized_string_with_extra_keys:
+    zero: Zero
```
kbairak, line 18: How is this discrepancy explained?

kbairak (approved):
Looks solid. Just the one question.

openformats/tests/formats/yamlinternationalization/files/1_en_exported.yml
sergafts: I was curious as well. I saw it in the input file and added it as well but the test defines plural rules as one, many. So zero is removed.

kbairak (approved):

PR #208, opened 2021-03-02
TX-12814 - Escaped character is not working properly with the STRUCTURED_JSON file format

arisktfx:
Related to: https://transifex.atlassian.net/browse/TX-12814 Escaped character is not working properly with the STRUCTURED_JSON file format

Problem and/or solution
-----------------------
When parsing a JSON or STRUCTURED JSON file the `_exctract` function is used to create the resource keys and get the resource string values. The resource key is created by recursively testing the value of the provided dictionary until it reaches a string. Then it combines all the preceding keys, joining them with a dot `.` .  For this to work with, if any of the keys has a dot in its name, it has to be escaped. E.g. given the JSON string:
```{
   "key1":{
      "key.2":{
         "key21":"a",
         "key22":"b",
         "key23":"c"
      },
      "key3":{
         "key31":"d",
         "key32":"e",
         "key33":"f"
      }
   }
}
```
the key for the first resource will be `key1.key\.2.key21` and the value `a`. Given a STRUCTURED JSON string like:
```
{
   "key1":{
      "key.2":{
         "string":"a",
         "comment":"b",
         "time":"c"
      },
      "key3":{
         "string":"d",
         "comment":"e",
         "time":"f"
      }
   }
}
```
the same logic was used to get the `string` value (e.g. key: `key1.key\.2.string`, value `a`) and then the "parent key" was extracted (e.g. `key1.key\.2.`) and used to retrieve the additional content for the resource (e.g. key: `key1.key\.2.comment`, value `b`). However, this would not work if the STRUCTURED JSON string was already escaped because the duplicated escaping would create a different key (e.g. `key1.key\\.2.string`)! Since the structure is know a different approach was used in the case of STRUCTURED JSON files, where the function recursively tries to find the first dictionary that has a key named "string". It then parses the data of this dictionary as a hole.

How to test
-----------
Parse a STRUCTURED JSON file with an escaped character in a key name, e.g `"Class:Location\/Attribute:geo"`

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/json.py
+++ b/openformats/formats/json.py
@@ -80,21 +80,41 @@ class JsonHandler(Handler):
                                      format(key, self.transcriber.line_number))
                 self.existing_keys.add(key)
 
-                if isinstance(value, (six.binary_type, six.text_type)):
-                    if not value.strip():
-                        continue
+                if self.name == "STRUCTURED_JSON":
+                    try:
+                        (string_value, _), = value.find_children(self.STRING_KEY)
+                    except:
+                        # Ignore other types of values like lists
+                        pass
+                    else:
+                        if string_value:
+                            if isinstance(string_value, (six.binary_type,
+                                                         six.text_type)):
+                                if string_value.strip():
+                                    openstring = self._create_openstring(
+                                        key, value)
 
-                    openstring = self._create_openstring(key, value,
-                                                         value_position)
-                    if openstring:
-                        self.stringset.append(openstring)
+                                    if openstring:
+                                        self.stringset.append(openstring)
 
-                elif isinstance(value, DumbJson):
-                    self._extract(value, key)
-
+                        elif isinstance(value, DumbJson):
+                            self._extract(value, key)
                 else:
-                    # Ignore other JSON types (bools, nulls, numbers)
-                    pass
+                    if isinstance(value, (six.binary_type, six.text_type)):
+                        if not value.strip():
+                            continue
+
+                        openstring = self._create_openstring(key, value,
+                                                             value_position)
+                        if openstring:
+                            self.stringset.append(openstring)
+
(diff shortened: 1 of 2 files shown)


igavriil (changes requested):
Great work for this tricky issue.

* It's not 100% clear to the reader when this issue happens and what's the reason behind it. Maybe mention why the STRUCTURED_JSON fails as opposed to KEYVALUEJSON and the backtracking that needs to happen from the `"string"` key to find the sibling attributes.
* I would prefer a more easy to follow naming for your variables. For example:
```
string_value_tuple = value.find_children(self.STRING_KEY)
if string_value_tuple[0][0]:
     ....
```
can be written as
```
(string, string_position), = value.find_children(self.STRING_KEY)
if string:
     ....
```
* There is a difference in the way additional properties are passed compared to the existing implementation. This is not easy to spot since the compiled file is not affected, but It could be affected in the future. Also it will create different metadata for existing strings which doesn't make much sense.

The source of that difference is that way that `DumbJson` handles the keys/values vs the traditional `json` module.
* `DumbJson` will parse the string exactly as is
* `json` module will parse the string and return the unescaped version of it.

What does this look like?
Let's take this json as an example: `{"foo": "bar\\baz"}`
* `DumbJson` will return the value as `"bar\\baz"`
* `json` will the return the value after un-escaping as  `"bar\baz"` (after all this is the string you wanted to have but JSON escaping rules need the `\` to be escaped as `\\`)

igavriil (changes requested):
Please provide a test case for the changed behavior.
Specifically it would be great if you could write a test that broke with the previous version but passes with the new one

igavriil (approved):
Great job. Not only fixing a broken behavior but also leaving behind a more readable code. Kudos

PR #209, opened 2021-03-05
Check hyperlinks after detecting text content

igavriil:
Problem and/or solution
-----------------------

We've identified an issue when compiling back a translation
file that was succesfully parsed. In more detail:

In the parsing phase all `w:p` (paragraph) elements are inspected.
For each paragraph we search for all `w:t` (text) elements
and extract the text that is wrapped inside this tag.
If text is found inside a `w:t` tag we also check to see if this
element is also related to a hyperlink by finding the `w:r` (run)
parent to see if it's type is `hyperlink`.

The above process runs fine on parsing but on compilation we also
do an extra step:
* We match the translation with the paragraph
* We wrap the translation into a fake tag in order to split it
  in text parts.
* For each `w:t` element we pop parts from the translations
* If there are more `w:t` elements that translation parts then we
  decompose (aka remove) from the document the part that exceed.

The above process not checking if a `w:t` was empty and the decomposition
took place before it should actually do, removing the `w:r` parent
when trying to found if a `w:t` was related to a hyperlink.

As an example we will use this document.xml

```
<w:body>
    <w:p w:rsidR="00000000" w:rsidDel="00000000" w:rsidP="00000000" w:rsidRDefault="00000000" w:rsidRPr="00000000" w14:paraId="000000F1">
        <w:pPr>
            <w:rPr/>
        </w:pPr>
        <w:r w:rsidDel="00000000" w:rsidR="00000000" w:rsidRPr="00000000">
            <w:rPr>
                <w:rtl w:val="0"/>
            </w:rPr>
            <w:t xml:space="preserve">Foo bar baz</w:t>
            <w:tab/>
            <w:t xml:space="preserve"></w:t>
            <w:tab/>
        </w:r>
    </w:p>
</w:body>
```
On parsing we inspect the paragraph `w:p` and all it's `w:t`
elements but the element that do not contain any text are
skipped. This way `<w:t xml:space="preserve"></w:t>` is ignored
and the output source string is just the contents of
`<w:t xml:space="preserve">Foo bar baz</w:t>` -> `Foo bar baz`

On compilation we get back the string `Φοο βαρ βαζ` we wrap in around
a pseudo tag `<wrapper>Φοο βαρ βαζ<wrapper>`
We iterate each `w:t` elements of the `w:p` and for each element we
find we pop out elements from the pseudo tag. So:

For `<w:t xml:space="preserve">Foo bar baz</w:t>` the string `Φοο βαρ βαζ`
is popped from the pseudo wrapper. Now the wrapper is empty.

Next comes `<w:t xml:space="preserve"></w:t>` but the wrapper is empty.
In such cases the parent (either hyperlink or w:p) is decomposed since
we don't have any text to replace.

Given this the above process we will only run only after we detect that `w:t`
actual has text.

Also from now on we will remove the `w:t` and not the `w:p` parent in cases
where the hyperlink is missing.

How to test
-----------
Added a test case for the specific docx.

Reviewer checklist
------------------

Code:

PR:


arisktfx (approved):
Looks good!

PR #221, opened 2021-07-14
Tx 13385 docx ampersand fix

kbairak:
Problem and/or solution
-----------------------

The issue is that BeautifulSoup does a "best effort" when converting its input to XML data. So, when a plain '&' is in the string, the information will be lost:

    >>> BeautifulSoup('<w>a & b</w>', 'xml').w.string
    <<< u'a  b'

This commit adds a `.replace('&', '&amp;')` to the constructor of the soup object used for translations to fix this issue

How to test
-----------

Add a '&' to a translation in the editor. Download the file in `devel` and in this branch.

Reviewer checklist
------------------

Code:

PR:


openformats/formats/docx.py
```diff
@@ -322,16 +329,22 @@ def compile(self, template, stringset, **kwargs):
...
+            # Do escaping: BeautifulSoup doesn't like unescaped '&' or '<' in
+            # its input. We do expect some '<tx>' and `</tx>` so we only escape
+            # '<' if it's not part of '<tx>'
+            translation = translation.replace(u"&", u"&amp;")
+            translation = re.sub(self.LT_PATTERN, u"&lt;", translation)
+
             translation_soup = BeautifulSoup(
-                u'<wrapper>{}</wrapper>'.format(translation), 'xml'
+                u'<wrapper>{}</wrapper>'.format(translation), 'xml',
             ).find_all(text=True)
```
igavriil, line 340: Instead of custom xml escaping, for some characters, I would prefer an approach like this:
```
from xml.sax.saxutils import escape

translation = translation.replace(u"<tx>", u"__OPENING_TAG__").replace(u"</tx>", u"__CLOSING_TAG__")
translation = escape(translation)
translation = translation.replace(u"__OPENING_TAG__", u"<tx>").replace(u"__CLOSING_TAG__", u"</tx>")
```

what do you think?

igavriil (commented):
Looks great!
I would prefer a different approach, since I am not a fun of regexes. Let me know what you think

openformats/formats/docx.py
kbairak: I like not having to use regex, but I don't like bringing in another library to help with escaping. I'll look into what can be done with using BeautifulSoup for escaping. If I can't figure out anything, I will use yours.

kbairak: It doesn't work because `<tx>` tags can have attributes. Even if I replace with `__OPENING_TAG__` only the `<tx` part, then the escape function will escape the closing `>` of the tag.

Reading the docs of `xml.sax.saxutils`, the `escape` function does very simple replacements. So we could keep your idea, replace only the `<tx` part and do the escaping with `.replace`, only replacing the `<` and `&` characters since apparently BeautifulSoup is ok with the `>` ones.

kbairak: done

codegaze (approved):

PR #223, opened 2021-08-13
Allow re-ordering of hyperlink tags

igavriil:
When a hyperlink is detected on the source file it will
always remain in the same place regarless if the translation
includes it in another part of the text.

With this commit the behavior is slightly altered:
Hyperlink tags can be:
* Removed: If the translation does not include a hyperlink in
  the specified location
* Replaced: If the translation includes an altered hyperlink in
  the specified location
* Added: If the translation includes a hyperlik in a location
  different than the source file

Problem and/or solution
-----------------------

How to test
-----------

Reviewer checklist
------------------

Code:

PR:


kbairak (approved):
Looks good, but needs rebase

kostasgeo (approved):

PR #224, opened 2021-08-18
Docx hyperlinks format

kostasgeo:
Problem and/or solution
-----------------------

Branch over https://github.com/transifex/openformats/pull/223
Releated Slack [thread](https://transifex.slack.com/archives/CPKBJ516F/p1628754343136200)

The solution in the above PR handles the re-ordering of hyperlink tags. As a result, when a hyperlink in the translation
of a docx string changes place, we manage to "track" it by deleting/adding hyperlinks in the source text, which is always our base.

In this PR we try to handle the last missing part, that is the tracking of the formatting of hyperlinks. When we add/remove hyperlinks above, although the downloaded translation file (compilation) has the correct hyperlinks, the formatting of the initial source file hyperlinks is kept, leading to highlighted text parts that are not hyperlinks, and new hyperlinks not highlighted.

Hyperlink formatting is insinde `w:rPr` tags as below:
```
<w:hyperlink>
<w:r w:rsidDel="00000000" w:rsidR="00000000" w:rsidRPr="00000000">
    <w:rPr>
        <w:color w:val="1155cc"/>
        <w:u w:val="single"/>
        <w:rtl w:val="0"/>
    </w:rPr>
    <w:t xml:space="preserve">
        Hyperlink
    </w:t>
</w:r>
</w:hyperlink>
```

The solution works as follows:
* While compiling the translation file, for each paragraph, we keep track of the added and deleted hyperlinks and the repsective text elements. This is done to cover the case where multiple hyperlinks need to be swapped with a different number of hyperlinks. e.g. a hyperlink on multiple strings: 
```
[
    u'one two ',
    u'<tx href="https://www.transifex.com/">three ',
    u'<tx>four</tx>',
    u'<tx> five </tx>',
    u'<tx>six</tx>',
    u'<tx> seven eight</tx></tx>',
    u'<tx> nine </tx>',
    u'<tx>ten </tx>',
    u'<tx>eleven</tx>',
    u' twelve'
]
``` 
like the above will result in multiple hyperlinks. If this is removed and we add the hyperlink to another place for a single text element, then we need to update the format of all those hyperlinks

* After we have kept track of the above, we go through the matched hyperlinks, added and deleted dicts and we swap only the 
rPr tags that seem to be related with the hyperlink formatting, that is the `<w:color>` and the the `<w:u>` (underline) tags.

How to test
-----------
Unit tests:
```
python setup.py test -s openformats.tests.formats.docx.test_docx.DocxTestCase.test_hyperlink_reorder
```

Manual testing:
* Reorder the hyperlinks in a translation, download the translation file and observe the correct format of the text in both removed and added hyperlinks. The actual tags swapping can be found in the `document.xml` file inside the zip.

Reviewer checklist
------------------

Code:

PR:


openformats/formats/docx.py
```diff
@@ -404,16 +408,54 @@ def compile(self, template, stringset, **kwargs):
                 # remove hyperlink from source docx
                 if hyperlink_url and not translation_hyperlink_url:
                     text_element.parent.parent.unwrap()
+                    if hyperlink_url in deleted_hl_text_elements.keys():
+                        deleted_hl_text_elements[hyperlink_url].append(text_element)
+                    else:
+                        deleted_hl_text_elements = {hyperlink_url: [text_element]}
```
igavriil, line 414: There is a handy wrapper for this built-in in python:

```
from collections import defaultdict
deleted_hl_text_elements = defaultdict(list)
```
Now we can append elements as all the keys are initialized as list (ie without need to check if key present)
```
deleted_hl_text_elements['foo'].append('bar') -> {'foo': ['bar']}
deleted_hl_text_elements['foo'].append('baz') -> {'foo': ['bar', 'baz']}
```

openformats/formats/docx.py
```diff
@@ -404,16 +408,54 @@ def compile(self, template, stringset, **kwargs):
...
+                if rpr.color:
+                    rpr.color.extract()
+                if color:
+                    rpr.append(color[0])
+                if rpr.u:
+                    rpr.u.extract()
+                if underline:
+                    rpr.append(underline[0])
+
+            if added_hl_text_elements and deleted_hl_text_elements and \
```
igavriil, line 443: Here we probably need to check that the we have as many hyperlinks added as removed.
We don't really care about the actual hyperlink, and since the key is the actual url this might cause issues.

For example it's possible `<href='www.transifex.com'>` to be translated as `<href='el.transifex.com'>` and this check will not hold, where it should.


So maybe better is:
```
if len(added_hl_text_elements) == len(deleted_hl_text_elements):

```

openformats/formats/docx.py
```diff
@@ -404,16 +408,54 @@ def compile(self, template, stringset, **kwargs):
...
+                added_hl_text_elements.keys() == deleted_hl_text_elements.keys():
+                for hl_url in added_hl_text_elements.keys():
```
igavriil, line 445: Following from above, here since the urls potentially will not match between source and translation, we need to combine in pairs the keys from each dictionary. (maybe with `zip`)?

openformats/formats/docx.py
kostasgeo: Good point! I missed that the actual URL might be translated as well.

PR #225, opened 2021-08-20
TX-12748 pot file not uploading for dz language

kostasgian:
Related to [TX-12748: pot file not uploading for dz language](https://transifex.atlassian.net/browse/TX-12748)

Problem and/or solution
-----------------------

If we create a resource with source language Dzongha and try to upload a po file with definitions for plurals, uploading will fail.

This happens because of the empty string in the msgstr[0] key. The PO handler expects not empty string, otherwise it needs both the single and plural rules in the language definition. 

The solution is to check the `msgstr_plural` and if it is has only one rule, produce only one `msgstr` string. 

How to test
-----------

* Use devel branch in the openformats.
* Create a new project with source language Dzongkha.
* Get the `api.po` file from the ticket and try to upload in order to create a new resource.
* Uploading the file you should get an error.
* Changing openformats to branch `TX-12748-pot-file-not-uploading-for-dz-language` the resource should be created.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ target/
 
 # Editor temp files
 .*.swp
+
+.idea
+*.iml
--- a/openformats/formats/po.py
+++ b/openformats/formats/po.py
@@ -223,10 +223,13 @@ class PoHandler(Handler):
         elif is_empty:
             if not self.only_values:
                 self.only_keys = True
-                string = entry.msgid if not pluralized else {
-                    0: entry.msgid,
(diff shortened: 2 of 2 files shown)


grass-tx (approved):
Looks fine. I'd propose the addition of a small unit test capturing exactly this problematic case with the empty string under `msgstr[0]` key, but I leave it up to you.

grass-tx (approved):

PR #226, opened 2021-08-20
Fix apping values that comprise only of an anchor

kbairak:
Problem and/or solution
-----------------------

The YaML parser is made to ignore anchors since they are irrelevant to localization and they are vulnerable to "billion laughs" attacks. So, when it encounters a:

```yaml
key: &anchor value
```

It will only consider the "value" part to be part of the extracted string. The problem is that, when the anchor **is** encountered, the
parser _expects_ there to be a value afterwards. When this is encountered:

```yaml
key: &anchor
```

according to the YAML spec, it should be equivalent to:

```yaml
key:
```

meaning the value of the key should be `null`.

```python
yaml.safe_load("key: &anchor")
# <<< {'key': None}
```

However, because the handler assumes that there is a value following the anchor, it does a `...split(' ', 1)[1]` which raises a `ValueError`.

The fix is it check against the captured YaMl node after an anchor is encountered and, if it's a null node, return it as-is.

How to test
-----------

Try uploading this file:

```yaml
key1: value1
key2: &anchor2 value2
key3: &anchor3
```

on `devel` and on this branch, which is what the brand-new unittest does anyway.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/yaml/utils.py
+++ b/openformats/formats/yaml/utils.py
@@ -81,15 +81,19 @@ class TxYamlLoader(yaml.SafeLoader):
         if anchor is None:
             return super(TxYamlLoader, self).compose_scalar_node(anchor)
         else:
-           node = super(TxYamlLoader, self).compose_scalar_node(anchor)
-           anchor_value = self.stream[
-               node.start_mark.index:node.end_mark.index
-            ].split(' ', 1)[1]
-           leading_spaces = len(anchor_value) - len(anchor_value.lstrip(' '))
+            node = super(TxYamlLoader, self).compose_scalar_node(anchor)
+            if node.tag == u'tag:yaml.org,2002:null':
+                # 'key: &anchor' should be interpreted as 'key:', ie the value
+                # should be ignored
+                return node
+            anchor_value = self.stream[
+                node.start_mark.index:node.end_mark.index
+             ].split(' ', 1)[1]
+            leading_spaces = len(anchor_value) - len(anchor_value.lstrip(' '))
 
-           node.start_mark.index = node.end_mark.index - len(anchor_value) \
-               + leading_spaces
-           return node
+            node.start_mark.index = node.end_mark.index - len(anchor_value) \
+                + leading_spaces
+            return node
 
     def _is_custom_tag(self, tag):
         """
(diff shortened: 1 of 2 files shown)


openformats/formats/yaml/utils.py
```diff
@@ -81,15 +81,19 @@ def compose_scalar_node(self, anchor):
...
-           anchor_value = self.stream[
-               node.start_mark.index:node.end_mark.index
-            ].split(' ', 1)[1]
-           leading_spaces = len(anchor_value) - len(anchor_value.lstrip(' '))
-
-           node.start_mark.index = node.end_mark.index - len(anchor_value) \
-               + leading_spaces
-           return node
+            node = super(TxYamlLoader, self).compose_scalar_node(anchor)
+            if node.tag == u'tag:yaml.org,2002:null':
```
kostasgian, line 85: I suppose this has an reasonable explanation, but since I am ignorant of what's going on under the hood, matching the node tag with this specific equation seems kind strange to me... 

So the tag is always coming as such when anchor value is null?

kostasgian: This is more a question to understand how this works... Tests are showing that indeed the solution works.

kbairak: That's a good question. TBH, I didn't check the docs to see what value to expect from `node.tag`. What I did was run the code through a debugger and inspect the value.

grass-tx: @kostasgian I think [this](https://yaml.org/spec/1.2/spec.html#id2805071) answers your question and also attests to the validity of @kbairak's findings.

deathbird: Browsing [PyYAML's docs](https://pyyaml.docsforge.com/master/documentation/#nodes) it seems a safe assumption.

deathbird (approved):
It works, Good job @kbairak!

openformats/formats/yaml/utils.py
kostasgian: Thanks @grass-tx and @deathbird for the clarifications!

kostasgian: I am adding here a small excerpt from the yaml specification, only for future reference. 
"If none of the regular expressions matches, the scalar is resolved to `tag:yaml.org,2002:str`"

kostasgian (approved):
LGTM

PR #239, opened 2021-10-15
Allow space control on docx files

igavriil:
Problem and/or solution
-----------------------

There are cases in docx files, where spaces
are wrapped in a text element on their own:
```
<w:r>
  <w:rPr/>
  <w:t xml:space="preserve"> </w:t>
</w:r>
```
In such cases in order to remove noise we do extract
the space surrounded by a `<tx>` tag (ie `<tx> </tx>`)
but rather the space is prepended to the next part of the string.

When compiling we always keep this empty text element
Regardless of the presence of this space in the translation
the existance of this empty text element is always adding
a space to the resulting document.

With this commit empty text elements are removed upon compilation
and we solely rely on the translation text for spaces.

How to test
-----------
A unit test in order to showcase that the spaces are controlled solely by the translation text.

Reviewer checklist
------------------

Code:

PR:


kostasgeo (approved):
LGTM

PR #245, opened 2021-11-15
TX-13671 - Skip `a:fld` tag

arisktfx:
Problem and/or solution
-----------------------
The `a:fld` tag is added when an automatically updated field (e.g. a page number) is added to the slides and usually contains no text to extract (it has an id), so I believe it is safe to let the code skip this tag and continue with the rest of the content

How to test
-----------
Uploading a pptx that contains an automatically updated field should be processed with no problem

Reviewer checklist
------------------

Code:

PR:


openformats/formats/office_open_xml/parser.py
```diff
@@ -90,9 +90,12 @@ def parse_paragraph(cls, paragraph, rels_soup):
                 text = u"".join([u" "*leading_spaces, text])
                 leading_spaces = 0
 
-            hyperlink_url = cls.get_hyperlink_url(
-                text_element, rels_soup
-            )
+            try:
```
manos-transifex, line 93: I don't know if it would make more sense to return None from get_hyperlink_url instead.

> Only concern is that this can hide any possible error.

igavriil (approved):

PR #267, opened 2021-12-17
Handle non Unicode spaces

arisktfx:
Problem and/or solution
-----------------------
Some Latin (ISO 8859-1) characters (e.g. `\xa0` : non-breakable space) cause the Github Markdown parser to skip the lines that contain them. By using `unicodedata.normalize` we convert these characters to their normal form.

How to test
-----------
Parsing an `.md` file that contains lines with characters like the above should be completed without skipping these lines.

Reviewer checklist
------------------

Code:

PR:


grass-tx (approved):
LGTM

PR #269, opened 2022-02-07
Problem with empty dicts when compiling

kbairak:
The KEYVALUEJSON compiler must remove sections of untranslated strings.
Thus, if you have a template like:

```json
{
    "key1": "hash1",
    "key2": "hash2"
}
```

and compile against a stringset that only has one of the two strings
translated, you should end up with a valid JSON file with only one of
the strings, like:

```json
{
    "key2": "translation2"
}
```

This gets a little more complicated with containers, like lists or
dicts. For example, assuming this template:

```json
{
    "key1": "hash1",
    "key3-4": ["hash3", "hash4"]
    "key4": "hash4"
}
```

If you compile against a stringset with translations for strings 1 and
4, then the whole list and its key should be removed:

```json
{
    "key1": "translation1",
    "key4": "translation4"
}
```

This is done by iterating over the container and, if no translations
were inserted during the iterations, get rid of the whole container.
This creates a problem when the container is empty in the template, this
logic generates a false-positive answer: "Yes, no hashes were replaced
with translations (because no hashes were there to begin with)". So, the
compiler attempts to remove the empty container from the compiled file,
with unexpected results.

This fix treats empty containers as ones where translations **were**
performed. ie, in response to the question:

- Did any replacements take place during the iteration?

Instead of answering:

- No, no replacements took place (none could have since the container
  was empty).

We now answer:

- Yes, all hashes in this container were replaced (technically true).

A test was also added to reproduce this issue.

Problem and/or solution
-----------------------

How to test
-----------

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/json.py
+++ b/openformats/formats/json.py
@@ -306,7 +306,7 @@ class JsonHandler(Handler):
         return at_least_one
 
     def _insert_from_dict(self, parsed, is_real_stringset):
-        at_least_one = False
+        at_least_one = not bool(list(parsed))
 
         for key, key_position, value, value_position in parsed:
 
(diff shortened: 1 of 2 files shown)


kostasgeo (approved):
Kudos, this was quite complex for me!

PR #281, opened 2022-08-05
Release: 0.0.102

wyngarde:
Merged Pull Requests:
#280 : fixing compliation when template value is null


manos-transifex (approved):
:rocket:

PR #282, opened 2022-08-17
Fix docx compilation error

petrosklg:
Related to [TX-14233](https://transifex.atlassian.net/browse/TX-14233).

Problem and/or solution
-----------------------

DOCX file fails to be exported due to compilation errors caused by empty tx tags.

There was an error raised when parent element does not exist.
This is fixed by adding a check to return if the parent is None. 

How to test
-----------

1. Create a new resource with the .docx file from the ticket.
2. Upload .xliff translation.
3. Click to download translated file.
4. It should not raise any errors.

Reviewer checklist
------------------

Code:

PR:


openformats/formats/docx.py
```diff
@@ -225,6 +225,9 @@ def remove_hyperlink(cls, text_element):
 
     @classmethod
     def remove_text_element(cls, text_element):
+        if text_element.find_parent('w:r') is None:
+            return
```
asgouros-tx, line 229: This is all out "of my league" here. Why don't we respond with e.g. `text_element.decompose()` instead? 
Should we ask for assistance from someone that is more familiar with openformats?

asgouros-tx (commented):
I did run the test-cases locally and indeed the fix seems to work.
However I am not very confident since it's the first time I see openformats.
I would propose that we seek a second reviewer from someone that is more apt to check the correctness of the fix here.

openformats/tests/formats/docx/test_docx.py
```diff
@@ -868,6 +868,43 @@ def test_lt(self):
...
+        self.assertEqual(openstring.string,
+                         u'This is a > greaterthan')
+        self.assertEqual(openstring.string, openstring.key)
+
+        # Compile with altered translation
+        translation = U'THIS IS A > GREATERTHAN'
+        stringset = [
+            OpenString(openstring.key, translation, order=0)
+        ]
+        content = self.handler.compile(template, stringset)
```
asgouros-tx, line 889: Is this case relevant to the issue here?

openformats/formats/docx.py
asgouros-tx: I tested compilation of the exact same file with the following method:
```
    @classmethod
    def remove_text_element(cls, text_element):
        if text_element.find_parent('w:r') is not None:
            run_parent = text_element.find_parent('w:r').parent
            if run_parent.name == 'hyperlink':
                return text_element.find_parent('w:hyperlink').decompose()

        return text_element.decompose()
```
and produced the same results. 
I am feeling more confident with this approach cause it still uses `text_element.decompose()`

openformats/tests/formats/docx/test_docx.py
asgouros-tx: I've gave it a check again. Test won't go into new condition imposed from https://github.com/transifex/openformats/pull/282/files#diff-475113b153a36064fc0fff2bdd9944c21a29e17e326364e12cdb9222bdfbda94R228. Please refactor so the change is covered.

openformats/formats/docx.py
petrosklg: Updated PR with proposed change.

openformats/tests/formats/docx/test_docx.py
asgouros-tx: Can you check this comment too?

asgouros-tx (changes requested):
Please also check comment about introduced test.

openformats/tests/formats/docx/test_docx.py
petrosklg: @asgouros-tx Tests updated.

asgouros-tx (approved):
:+1:

PR #289, opened 2022-10-13
Fix STRUCTURED_JSON compilation when file includes empty strings

igavriil:
Problem and/or solution
-----------------------
When an empty string was detected in the source file this string was not included in the stringset.
Later at compile time even if the string was empty the stringset iterator was forwarded so the resulting
file was corrupted.

From now we detect such conditions and we handle them correctly.

How to test
-----------
A simple unit test was added to illustrate this scenario.

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/json.py
+++ b/openformats/formats/json.py
@@ -9,12 +9,12 @@ from itertools import count
 
 import six
 
-from ..exceptions import ParseError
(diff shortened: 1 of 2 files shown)


manos-transifex (approved):
:godmode:

PR #299, opened 2023-04-24
Wrong parsing of paragraphs with bullet and url

arisktfx:
Problem and/or solution
-----------------------
Parsing slides with bullets that contained urls was creating wrong number of tx tags. If these strings was then translated using an automated proccess (eg Machine translation) the returned translation would be wrongly formated resulting in a corrupt .pptx file when downloaded.

How to test
-----------
Upload a pptx file that contains a slides with paragraphs with bullets and a url, machine translate the file and then `Download for use` the file, PowerPoint should not warn of corrupt file.

Reviewer checklist
------------------

Code:

PR:


arisktfx:
Ticket will be handled in a different way

PR #305, opened 2023-08-03
Structured json parser: Hashes displayed on compiled template

foteinigk:
If for some reason iterator of translations returns None, the hash remains at compiled translation We can skip source string in this case

Problem and/or solution
-----------------------

How to test
-----------

Reviewer checklist
------------------

Code:

PR:


openformats/formats/json.py
```diff
@@ -567,7 +567,13 @@ def _compile_recursively(self, current_part):
                         skip=True
                     else:
                         translation = next(self.translations, None)
-                        skip=False
+                        if translation is None:
```
deathbird, line 570: I don't object to your addition. Since however it's an edge case (that shouldn't normally happen) and also hard to reproduce let's document here roughly when this braching logic you introduce happens. Also please add a respective test in `openformats/openformats/tests/formats/structuredkeyvaluejson/test_keyvaluejson.py` that covers the this case.

deathbird (changes requested):

openformats/formats/json.py
deathbird: Sorry about that but after consideration I have second thoughts on this @foteinigk...

Current state: both `StructuredJsonHandler` and `ChromeI18nHandler` (ver1 + ver2) leave the hash in the template while `JsonHandler` and `ChromeI18nHandlerv3` remove the hash from the template. With the change you'll introduce we will have a thrid (different) behaviour (given that `ChromeI18nHandler` ver1 + ver2 is used only for old chrome json files - new one's use the v3 version of the handler which does what ). Won't this introduce confusion? To tell you the truth I'm leaning more to the behaviour that `JsonHandler` and `ChromeI18nHandlerv3` exhibit... What do you think?

foteinigk: Yeh better keep the existing functionality for consistency👍

PR #325, opened 2024-02-01
fix corrupted google sheet

manos-transifex:
Problem and/or solution
-----------------------

When we compile an xlsx we should not decomponese a t tag (but we should set it to empty string)
The reason is that if we remove the <t> tag we should also remove the related <rPr> otherwise gsheets cannot open the sheet.

this is an example of a valid string object in sharedStrings.xml
```
    <si>
        <r>
            <rPr>
                <rFont val="Open Sans" />
                <b />
                <color theme="1" />
            </rPr>
            <t xml:space="preserve"></t>
        </r>
        <r>
            <rPr>
                <rFont val="Open Sans" />
                <b />
                <color theme="1" />
            </rPr>
            <t xml:space="preserve">string2</t>
        </r>
    </si>
````
when we parse the string we
1) extract non empty <t> elements 
2) wrap the strings in <tx></tx> tags to preserve formatting and return the result without adding unused empty tags
3) when we compile the document we remove add replace the translated strings to the first non empty <t> tags instead

before when we replacing the translation parts we were decomposing (destruct) the <t> tags which was leaving garbage like
```
<si>
        <r>
            <rPr>
                <rFont val="Open Sans" />
                <b />
                <color theme="1" />
            </rPr>
            <t xml:space="preserve">string2</t>
        </r>
        <r>
            <rPr>
                <rFont val="Open Sans" />
                <b />
                <color theme="1" />
            </rPr>
        </r>
    </si>
```
the above is invalid for google sheets
the valid alternatives are either
1) remove the whole <r> element or 
2) simply set the <t> tag string to an empty string

How to test
-----------

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/xlsx_unstructured.py
+++ b/openformats/formats/xlsx_unstructured.py
@@ -10,10 +10,21 @@ from uuid import uuid4
 from zipfile import ZipFile, ZIP_DEFLATED
 
 from bs4 import BeautifulSoup
+from bs4.dammit import EntitySubstitution
+from bs4.formatter import XMLFormatter
 
 from openformats.formats.office_open_xml.parser import OfficeOpenXmlHandler
 from openformats.handlers import Handler
 from openformats.strings import OpenString
+
+
+class UnsortedAttributes(XMLFormatter):
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


fathineos (approved):
Nice work and verbose description!

PR #333, opened 2024-04-25
Accept newlines and tabs in android v3

kbairak:
Problem and/or solution
-----------------------

Some previous improvements were a bit overzealous. We were supposed to reject some characters that are not supported by Android and/or XML. We rejected newline and tab characters along with this. This created problems for the customer since the Android SDK tools accept newline characters in the strings

How to test
-----------

Set up transifex to use this parser and upload this file:

```xml
<resources>
   <string name="with newline">
        Free crypto
    </string>
    <string name="with escaped newline">Free \n crypto</string>
</resources>
```

Both uploads should work.

The first string should appear like this in the editor in rich mode:

image

and like this in raw mode:

image

And the second string should look like this in rich mode:

image

and like this in raw mode:

image

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/android_unescaped.py
+++ b/openformats/formats/android_unescaped.py
@@ -3,8 +3,6 @@ from hashlib import md5
 from openformats.exceptions import ParseError
 from openformats.formats.android import AndroidHandler
 from ..utils.xml import NewDumbXml as DumbXml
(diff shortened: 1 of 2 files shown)


petrosklg (approved):
lgtm

PR #337, opened 2024-05-17
Adding support for WebVTT (Web Video Text Tracks) (.vtt) format

dsavinov-actionengine:
Problem and/or solution
-----------------------
Adding parsing and compiling for WebVTT (Web Video Text Tracks) (.vtt) format

How to test
-----------
### 1. Running unit-tests
`/openformats/tests/formats/vtt/test_vtt.py` contains tests for vtt
Use `pytest openformats/tests/formats/vtt/test_vtt.py` to run tests

### 2. Through testbed
Use "VTT" handler in the testbed

Reviewer checklist
------------------

Code:

PR:


dsavinov-actionengine:
@kbairak please review

openformats/formats/vtt.py
```diff
@@ -0,0 +1,174 @@
...
+                f"Problem with end of timing at line {self.transcriber.line_number + 1}: '{end}'"
+            )
+
+        # Content
+        string_to_translate = '\n'.join(src_strings[timings_index+1:])
+        if string_to_translate == "":
+            raise ParseError(f"Subtitle is empty on line {self.transcriber.line_number + 2}")
+
+        string = OpenString(timings, string_to_translate,
+                            occurrences=f"{start},{end}")
```
kbairak, line 104: It's good practice to include an `order` with each OpenString. This is somewhat easy to do:

```python
from itertools import count

order = count()
for ... in ...:
    string = OpenString(..., order=next(order))
```

This will ensure each string will get an auto-incrementing value for `order`.

kbairak (commented):
Overall it looks good. I will take a more thorough look and invoke it through a debugger. I have a feeling some things can be simplified but I will have to experiment a bit. For the time being, just the one comment.

openformats/formats/vtt.py
```diff
@@ -0,0 +1,177 @@
...
+    def _parse_section(self, offset, section):
+        src_strings = section.split('\n')  # identifier_str is optional in VTT
+
+        timings = ""
+        timings_index = -1
+        for i in range(len(src_strings)):
+            str = src_strings[i];
+            if "-->" in str:
+                timings = str
+                timings_index = i
```
kbairak, line 60: Minor because it is not going to affect performance by a lot, but we could `break` here.

Actually, we _should_ break because the `-->` part could be in an actual subtitle and we want to consider the first occurrence as the timing.

openformats/formats/vtt.py
```diff
@@ -0,0 +1,177 @@
...
+
+        string = OpenString(timings, string_to_translate,
+                            occurrences=f"{start},{end}",
+                            order=next(self._order))
+        offset += len(identifier) + len(timings) + 1;
+        if len(identifier):
+            offset += 1
+        return offset, string
+
+    def _format_timing(self, timing):
```
kbairak, line 113: This is part of the standard library. It might be more well-suited. https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

openformats/formats/vtt.py
```diff
@@ -0,0 +1,177 @@
...
+                                                    int(milliseconds))
+            return f"{hours:02}:{minutes:02}:{seconds:02}.{milliseconds:03}"
+        else:
+            raise ParseError(f"Unexpected timing format on line {self.transcriber.line_number + 2}")
+
+    def compile(self, template, stringset, **kwargs):
+        transcriber = Transcriber(template)
+        template = transcriber.source
+        stringset = iter(stringset)
+        string = next(stringset)
```
kbairak, line 140: There is a small probability (and if we're honest, transifex will probably stop the compilation process before the interpreter gets here) that this will raise a `StopIteration`. Maybe a try/except that raises a `ParseError("stringset cannot be empty")` would fit here.

openformats/formats/vtt.py
```diff
@@ -0,0 +1,177 @@
...
+        for start, subtitle_section in self._generate_split_subtitles(template):
+            transcriber.copy_until(start)
+            transcriber.mark_section_start()
+
+            # Find hash after timings
+            hash_position = -1
+            if subtitle_section.count('-->') > 0:
+                arrow_pos = subtitle_section.index('-->')
+                end_of_timings = subtitle_section.index('\n', arrow_pos + len('-->'))
+                hash_position = end_of_timings + 1
```
kbairak, line 151: How about:

```suggestion
            try:
                arrow_pos = subtitle_section.index('-->')
            except ValueError:
                hash_position = -1
            else:
                end_of_timings = subtitle_section.index('\n', arrow_pos + len('-->'))
                hash_position = end_of_timings + 1
```

I know mine is longer and this is a styling preference so feel free to ignore but I feel it is more _pythonic_.

openformats/formats/vtt.py
```diff
@@ -0,0 +1,177 @@
...
+
+        for start, subtitle_section in self._generate_split_subtitles(template):
+            transcriber.copy_until(start)
+            transcriber.mark_section_start()
+
+            # Find hash after timings
+            hash_position = -1
+            if subtitle_section.count('-->') > 0:
+                arrow_pos = subtitle_section.index('-->')
+                end_of_timings = subtitle_section.index('\n', arrow_pos + len('-->'))
```
kbairak, line 150: Could this line raise a `ValueError`?

kbairak (changes requested):
Great work! I think that only the first comment _requires_ a change (the one with the multiple occurrences of `-->`). The rest are optional.

openformats/formats/vtt.py
dsavinov-actionengine: Done

openformats/formats/vtt.py
dsavinov-actionengine: This function is written based on a similar function in SRT handler. It takes a string as input and returns a string too. The built-in strftime()/strptime() are less convenient here. Let's leave this function as it is?

openformats/formats/vtt.py
dsavinov-actionengine: Done

openformats/formats/vtt.py
dsavinov-actionengine: In current PR code, `.index('-->')` cannot produce an exception because it is under condition `(.count('-->') > 0)` in the previous line. 
But `.index('\n', arrow_pos + len('-->'))`  (in both current code and in your suggestion) can give an exception.
So in the new commit there is a rework.

kbairak (approved):
Great work!

PR #338, opened 2024-05-24
Adding support for ARB (Application Resource Bundle) (.arb) format

dsavinov-actionengine:
Problem and/or solution
-----------------------
Adding parsing and compiling for ARB (Application Resource Bundle) (.arb) format

How to test
-----------
### 1. Running unit-tests
`/openformats/tests/formats/arb/test_arb.py` contains tests for arb
Use `pytest /openformats/tests/formats/arb/test_arb.py` to run tests

### 2. Through testbed
Use "ARB" handler in the testbed

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/json.py
+++ b/openformats/formats/json.py
@@ -162,7 +162,7 @@ class JsonHandler(Handler):
         # e.g. a pluralized string.
         # If it cannot be parsed that way (returns None), parse it like
         # a regular string.
-        parser = ICUParser(allow_numeric_plural_values = (self.name == "ARB"))
+        parser = ICUParser(allow_numeric_plural_values = False)
         icu_string = parser.parse(key, value)
         if icu_string:
             return self._create_pluralized_string(icu_string, value_position)
@@ -171,7 +171,8 @@ class JsonHandler(Handler):
             key, value, value_position
         )
 
-    def _create_pluralized_string(self, icu_string, value_position):
+    def _create_pluralized_string(self, icu_string, value_position,
+                                  context_value="", description_value=""):
         """Create a pluralized string based on the given information.
 
         Also updates the transcriber accordingly.
@@ -185,6 +186,8 @@ class JsonHandler(Handler):
             icu_string.strings_by_rule,
             pluralized=icu_string.pluralized,
             order=next(self._order),
+            context=context_value,
+            developer_comment=description_value,
         )
 
         current_pos = icu_string.current_position
@@ -196,7 +199,8 @@ class JsonHandler(Handler):
 
         return openstring
 
-    def _create_regular_string(self, key, value, value_position):
+    def _create_regular_string(self, key, value, value_position,
+                               context_value="", description_value=""):
         """Return a new simple OpenString based on the given key and value
         and update the transcriber accordingly.
 
@@ -204,7 +208,10 @@ class JsonHandler(Handler):
         :param value: the translation string
         :return: an OpenString or None
         """
-        openstring = OpenString(key, value, order=next(self._order))
+        openstring = OpenString(key, value, order=next(self._order),
+                                context=context_value,
+                                developer_comment=description_value,
+        )
         self.transcriber.copy_until(value_position)
         self.transcriber.add(openstring.template_replacement)
         self.transcriber.skip(len(value))
@@ -485,11 +492,14 @@ class ArbHandler(JsonHandler):
         source = self.transcriber.source
         self.stringset = []
         self.existing_keys = set()
+        self.metadata = dict()
 
         try:
             parsed = DumbJson(source)
         except ValueError as e:
             raise ParseError(six.text_type(e))
+        if parsed.type != dict:
+            raise ParseError("Invalid JSON")
         self._order = count()
         self._find_keys(parsed)
         self._extract(parsed)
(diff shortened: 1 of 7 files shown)


dsavinov-actionengine:
Tagging @kbairak for review, but feel free to tag others if needed

openformats/formats/json.py
```diff
@@ -472,6 +472,123 @@ def unescape(string):
...
+                    self.transcriber.copy_until(key_position)
+                    raise ParseError(u"Duplicate string key ('{}') in line {}".
+                                     format(key, self.transcriber.line_number))
+                if nest is None:  # store all root-level keys in order to detect duplication
+                    self.existing_keys.add(key)
+                elif key.startswith("@") and key.endswith(".type") and value != "text":
+                    self.existing_keys.add(key)
+
+                if isinstance(value, DumbJson):
+                    self._find_keys(value, key)
```
kbairak, line 523: Instead of going infinitely deep, we could instead search up to level two and collect entries that are non-text. This way we don't have to bother with escaping keys etc. Here is an idea; I haven't tested it so there might be a few errors:

```python
class ArbHandler(JsonHandler):
    ...

    def parse(self, content, **kwargs):
        ...

        keys = self._find_keys(parsed)
        stringset = self._extract(parsed, keys)

    def _find_keys(self, parsed):
        if parsed.type != dict:
            raise ParseError("...")

        keys, keys_to_ignore = set(), set()
        for key, key_position, value, _ in parsed:
            if not key.startswith("@"):
                if not isinstance(value, str):
                    continue
                if key in keys:
                    raise ParseError(
                        f"Duplicate string key ('{key}') in line "
                        f"{self.transcriber.line_number}"
                    )
                keys.add(key)
            else:
                if isinstance(value, DumbJson):
                    (inner_value, _), = value.find_children('type')
                    if inner_value == "text":
                        keys_to_ignore.add(key)
        return keys - keys_to_ignore

    def _extract(self, parsed, keys):
        stringset = []
        for key, _, value, value_position in parsed:
            if key not in keys:
                continue
            ...
        return stringset
```

openformats/formats/json.py
```diff
@@ -472,6 +472,123 @@ def unescape(string):
...
+                                                         value_position)
+                    if openstring:
+                        self.stringset.append(openstring)
+                else:
+                    # Ignore other JSON types (bools, nulls, numbers)
+                    pass
+        else:
+            raise ParseError("Invalid JSON")
+
+    def compile(self, template, stringset, **kwargs):
```
kbairak, line 553: So, the only reason you re-define `compile` and `_copy_until_and_remove_section` is so that you can add the `keep_sections` functionality, which is only needed for testing. I am not very fond of this idea. It makes the code unnecessarily complex. However, there is a case where not removing the sections makes sense.

This is not adequately described in the docs, but when Transifex invokes this handler, it will pass `is_source` as a keyword argument. In this case, we can be certain that

1. the stringset will not be missing anything and that
2. the user wants to get the exact same file they previously uploaded

So, if this is indeed the same use-case, maybe you can rename `keep_sections` to `is_source`.

In any case, lets make sure that when `is_source=True` is passed to `compile`, the end result is the same as the originally uploaded file.

openformats/formats/json.py
```diff
@@ -472,6 +472,123 @@ def unescape(string):
...
+
+                if isinstance(value, DumbJson):
+                    self._find_keys(value, key)
+                else:
+                    pass
+        else:
+            raise ParseError("Invalid JSON")
+
+    def _extract(self, parsed):
+        if parsed.type == dict:
```
kbairak, line 530: No need for this check, we already did it in `_find_keys`. Perhaps to make it clearer, lets make one check in `parse` and then in `_find_keys` and `_extract` assume that `parsed` represents a dict.

kbairak (changes requested):
Great job!

kbairak:
There is another issue I am concerned about. Looking at the examples of the [official Flutter documentation](https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#setting-up), it looks like the language files are supposed to be more bare: only key-translation pairs without any metadata.

image

So, the question is: should we follow this? should we make sure that when `is_source=False` during `compile`, only the key-translation pairs should be present while when `is_source=True` we should make sure that they get back the same file they originally uploaded?

Have you looked into this?

openformats/formats/json.py
dsavinov-actionengine: Did not try this. Anyway, ARBs normally have only two levels of json, so we don't expect infinite nesting.

openformats/formats/json.py
dsavinov-actionengine: Tried this, replaced `keep_sections` with `is_source`, but that did not work, unit-tests failed. Anyway, re-definition of the `compile()` method is necessary now, as it handles language code for ARB.

openformats/formats/json.py
dsavinov-actionengine: done

dsavinov-actionengine:
> There is another issue I am concerned about. Looking at the examples of the [official Flutter documentation](https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#setting-up)

...
>should we make sure that when is_source=False during compile, only the key-translation pairs should be present while when is_source=True we should make sure that they get back the same file they originally uploaded?
> Have you looked into this?

Did not implement this logic in the latest commit. It looks unusual if compared to other openformats handlers, which have their compiled files as similar as possible to the source files (only strings from stringset change). Maybe we could try to get back to this idea in a separate pull-request.

openformats/formats/json.py
```diff
@@ -472,6 +479,156 @@ def unescape(string):
...
+            if key.startswith("@"):
+                continue
+            elif isinstance(value, (six.text_type)):
+                if not value.strip():
+                    continue
+                elif f"@{key}.type" in self.existing_keys:
+                    continue
+
+                context_key = f"@{key}.context"
+                context_value = self.metadata[context_key] \
```
dsavinov-actionengine, line 554: Implemented issue 1 from QA spreadsheet

openformats/formats/json.py
```diff
@@ -472,6 +479,156 @@ def unescape(string):
...
+                    order=openstring.order,
+                    pluralized=openstring.pluralized)
+            for openstring in stringset
+        ]
+        new_template = self._replace_translations(
+            template, fake_stringset, False
+        )
+        new_template = self._clean_empties(new_template)
+
+        if language_info is not None:
```
dsavinov-actionengine, line 609: Implemented issue 2 from QA spreadsheet

openformats/formats/json.py
```diff
@@ -472,6 +479,156 @@ def unescape(string):
...
+                if not value.strip():
+                    continue
+                elif f"@{key}.type" in self.existing_keys:
+                    continue
+
+                context_key = f"@{key}.context"
+                context_value = self.metadata[context_key] \
+                    if context_key in self.metadata.keys() else ""
+                description_key = f"@{key}.description"
+                description_value = self.metadata[description_key] \
```
dsavinov-actionengine, line 557: Implemented issue 3 from QA spreadsheet

kbairak (approved):

PR #341, opened 2024-06-24
Fix VTT issues

dsavinov-actionengine:
Problem and/or solution
-----------------------
VTT fixes for issues found during the 1st QA pass

How to test
-----------

Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/vtt.py
+++ b/openformats/formats/vtt.py
@@ -45,7 +45,9 @@ class VttHandler(Handler):
         self.transcriber.copy_until(len(source))
 
         template = self.transcriber.get_destination()
-        if not template.startswith("WEBVTT"):
+        if len(stringset) == 0:
+            raise ParseError("There are no strings to translate")
+        elif not template.startswith("WEBVTT"):
             raise ParseError("VTT file should start with 'WEBVTT'!")
         return template, stringset
 
--- a/openformats/formats/vtt.py
+++ b/openformats/formats/vtt.py
@@ -45,7 +45,9 @@ class VttHandler(Handler):
         self.transcriber.copy_until(len(source))
 
         template = self.transcriber.get_destination()
-        if not template.startswith("WEBVTT"):
+        if len(stringset) == 0:
+            raise ParseError("There are no strings to translate")
+        elif not template.startswith("WEBVTT"):
             raise ParseError("VTT file should start with 'WEBVTT'!")
         return template, stringset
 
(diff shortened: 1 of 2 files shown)


openformats/formats/vtt.py
```diff
@@ -45,7 +45,9 @@ def parse(self, content, **kwargs):
         self.transcriber.copy_until(len(source))
 
         template = self.transcriber.get_destination()
-        if not template.startswith("WEBVTT"):
+        if len(stringset) == 0:
+            raise ParseError("There are no strings to translate")
```
dsavinov-actionengine, line 49: Implemented issue 6 from QA spreadsheet: system should fail if source content consists of only WEBVTT signature

openformats/formats/vtt.py
```diff
@@ -101,10 +103,6 @@ def _parse_section(self, offset, section):
 
         # Content
         string_to_translate = "\n".join(src_strings[timings_index + 1 :])
-        if string_to_translate == "":
-            raise ParseError(
-                f"Subtitle is empty on line {self.transcriber.line_number + 2}"
```
dsavinov-actionengine, line 106: Implemented issue 3 from QA spreadsheet: system should allow uploading (parsing) even if some subtitles are empty

dsavinov-actionengine:
@kbairak 
**Important note below for issue 1** ("when the stringset sent to the compiler has no strings, the source strings should be added instead to that stringset.")

**In short:** At openformats level, we don't have info during `compile()` about the source strings. The fix might be appliet in transifex code.
**Details:**
When Transifex calls `parse` method, the `parse(content)` makes a template (where hashes replace strings) and a stringset from the `content`. 
Then, when Transifex calls our `compile(template, stringset)` method, **only the new** translated stringset is passed to us, but we don't get the original stringset or at least the original `content` in the `compile()` method. It seems only Transifex has access to everything at once: source content, source stringset, template, translated stringset.

Please let us know your thoughts on this.

dsavinov-actionengine:
@kbairak added fixes for AE-45 + checked for scenarios you mentioned:

1. during parsing, if the string that's about to be extracted is empty, don't include it in the resulting stringset at all
2. during compiling, if the stringset has an empty string, the resulting file should have this string's section but with the string content empty
3. during compiling, if a string is missing from the stringset, the resulting file should not have the string's section at all (the timestampts and anything else should be removed)

Testable were 1 and 2. 3 is only testable outside openformats testbed

kbairak:
Hey @dsavinov-actionengine, thanks for the update. I will start checking.

This

> 3 is only testable outside openformats testbed

however, is not true. You can do/write a test like this:

```python
handler = VttHandler()
template, stringset = handler.parse(...)  # A source text with 3 strings
compiled = handler.compile(template, stringset[:2])  # skip the 3rd string
assertEqual(compiled, ...)
```

kbairak (approved):

PR #349, opened 2024-07-29
Android parser: Fix for escaping  character <

foteinigk:
Problem and/or solution
-----------------------
Currently, when calling AI/MT services to translate a string, we check if the resource supports raw formatting. If it does, we unescape the strings to ensure that "rich" mode is always passed to the AI/MT services. After receiving the translation, we escape the translated strings before storing them in the database.

However, for strings like` < 20 units`, the escape function fails due to a parsing error. Specifically, the Android parser tries to parse the uploaded content and detect inline protected tags, but this fails with an error indicating the absence of a closing `> `tag.

To address this, we can catch the exception during the parsing of inline tags and escape all special characters  the string. 

How to test
-----------
Source string => String in translations downloaded file
 '< 20 units'  =>  '&lt; 20 units'
 '< 20 & > 50 units<xliff:g>test</xliff:g>'' => the xlif protected tag is not respected in this case. Unfortunately, i was not able to find a solution to handle this case as well
       
Reviewer checklist
------------------

Code:

PR:

### DIFF
--- a/openformats/formats/android_unescaped.py
+++ b/openformats/formats/android_unescaped.py
@@ -4,7 +4,6 @@ from openformats.exceptions import ParseError
 from openformats.formats.android import AndroidHandler
 from ..utils.xml import NewDumbXml as DumbXml
 
-
 class AndroidUnescapedHandler(AndroidHandler):
     def _create_string(self, name, text, comment, product, child, pluralized=False):
         """Creates a string and returns it. If empty string it returns None.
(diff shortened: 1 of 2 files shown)


openformats/formats/android_unescaped.py
```diff
@@ -96,7 +96,21 @@ def escape(string):
...
+            string = AndroidHandler.escape(string)
+            string = (
+                string.replace("&", "&amp;")
+                .replace(">", "&gt;")
+                .replace("<", "&lt;")
+                .replace("\n", "\\n")
+                .replace("\t", "\\t")
+                .replace("@", "\\@")
+                .replace("?", "\\?")
+            )
```
asgouros-tx, line 112: I noticed we have almost an exact handling of these special characters (with the exception of `<` character) in lines [116-123](https://github.com/transifex/openformats/pull/349/files#diff-b74cd526265af6c90df41083c31dbe304f518d474b2c3bdb2d22bf87ff6b127cL102-L109). Can we avoid the duplication (and possibly also "fix" the fact that replacement of `<` is missing)?

foteinigk: Sure, we can create a function for the escaped characters that are common in both cases. Just a small note, I initially thought the '<' was missing from the escaping characters, but it appears this was intentional to avoid escaping '<' within protected inline tags. I'll add a comment to clarify this and avoid any confusion.🙏

foteinigk: i have pushed the fixes🙏

asgouros-tx (approved):

PR #392, opened 2026-04-30
Fix GithubMarkdownHandlerV2 fences parsing

deathbird:
Problem and/or solution
-----------------------
Introduce a new handler that fixes two CommonMark issues in
mistune 0.8 without changing the behaviour of the existing V1
or V2 handlers (kept verbatim for backwards compatibility with
already-uploaded customer files):

1. Fence info strings. Mistune's fence regex only accepts
   `^ *(\`{3,}|~{3,}) *(\S+)? *\n`, so a CommonMark-legal opener
   like ` ```js lines ` fails to match, the block falls into the
   paragraph rule, and the paragraph keeps consuming through the
   following heading. Fixed via TxBlockGrammarV2_1 / TxBlockLexerV2_1,
   which override the fences pattern and rebuild paragraph against
   it (paragraph embeds the fences regex in a lookahead).

2. Introduce a new MDX handler that picks up the fence info-string
fix and the inline code-span split from GithubMarkdownHandlerV2_1.
The original MarkdownJsxHandler is left untouched (still inherits
from GithubMarkdownHandlerV2) so existing customer .mdx resources
keep parsing through the exact same code path and templates do not
shift under their feet.

How to test
-----------
Upload a .md file containing a fenced code block like:
````
```js filename="xxx.js"
exports.onExecutePostLogin = async (event, api) => {
  if (event.client.name === "APP_NAME") {
    const d = new Date().getDay();

    if (d === 0 || d === 6) {
      api.access.deny("This app is only available during the week.");
    }
  }
}
```

````
followed by other markdown content (eg paragraphs) before and after the fix. Before the fix the above code block is broken in more than one strings on the stringset.

Reviewer checklist
------------------

Code:

PR:

### CODE REVIEW

File: openformats/formats/github_markdown_v2_1.py [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+            for openstring in yaml_stringset:
+                self._unescape_non_printable(openstring)
+        else:
+            md_content = content
+
+        md_template = md_content
+
+        # Use the V2_1 lexer so the permissive fences/paragraph grammar kicks
+        # in. Everything else is identical to GithubMarkdownHandlerV2.parse.
+        block = TxBlockLexerV2_1()
```
arisktfx: If I am not mistaken all the parse function is a copy from GithubMarkdownHandlerV2 except from this line, so why not just create a new method in the parent class eg:
```
def _get_block_lexer(self):
    return TxBlockLexer()
```
and in parent change to:
```
block = self._make_block_lexer()
```
and then just override the new method to GithubMarkdownHandlerV2_1
```
def _get_block_lexer(self):
    return TxBlockLexerV2_1()
```

arisktfx (commented):
Logic looks fine, just left a suggestion to keep it DRY

arisktfx (approved):
✔️

